PR #27, opened 2021-03-03
Correct license; add make support to add license

ChunyiLyu:
## Summary Of Changes

Correct license; `make generate` should attach the correct license to generated files now too


MirahImage (approved):
LGTM

PR #28, opened 2021-03-04
Do not requeue error if queue/exchanges are already deleted

### LINKED ISSUE #26: Queue resource deletion when the queue has already been deleted within RMQ
When a queue has already been deleted within RMQ, deleting the queue resource fails. Instead the deletion should be considered successful. 

When deleting a queue resource and the corresponding queue has already been deleted in RMQ, when the controller receives the failure message from RMQ that the object does not exist, the deletion should be considered successful and proceed. 

The controller could check if the object exists prior to attempting to delete, but this would increase the number of API calls in the happy path, impacting performance.

ChunyiLyu:
This closes #26

## Summary Of Changes

- ignore '404' NotFound error from server when deleting exchanges and queues
- autoDelete queue and exchanges could be autodeleted from server when the last consumer unsubscribe; should not requeue on error when the object is already gone

### DIFF
--- a/controllers/queue_controller.go
+++ b/controllers/queue_controller.go
@@ -12,6 +12,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"github.com/go-logr/logr"
 	rabbithole "github.com/michaelklishin/rabbit-hole/v2"
 	topologyv1alpha1 "github.com/rabbitmq/messaging-topology-operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


controllers/queue_controller.go
```diff
@@ -120,10 +120,16 @@ func (r *QueueReconciler) addFinalizerIfNeeded(ctx context.Context, q *topologyv
...
 
+// deletes queue from rabbitmq server
+// if server responds with '404' Not Found, it logs and does not requeue on error
+// queues could be deleted manually or gone because of AutoDelete
 func (r *QueueReconciler) deleteQueue(ctx context.Context, client *rabbithole.Client, q *topologyv1alpha1.Queue) error {
 	logger := ctrl.LoggerFrom(ctx)
 
-	if err := validateResponse(client.DeleteQueue(q.Spec.Vhost, q.Spec.Name)); err != nil {
+	err := validateResponseForDeletion(client.DeleteQueue(q.Spec.Vhost, q.Spec.Name))
+	if err == NotFound {
```
coro, line 130: My preferred method for checking error types in Go is with 
```go
errors.Is(err, NotFound)
```
As it checks the whole call stack of potentially wrapped errors. It doesn't matter for this particular case, since you're only creating one level of errors, so up to you if you want to bother.

coro (approved):
LGTM

controllers/queue_controller.go
ChunyiLyu: Done!

PR #29, opened 2021-03-04
Allow creation & deletion of users with the operator

### LINKED ISSUE #3: Use the operator to create users

- crd name: users.rabbitmq.com
- all configurations
- follow the [provision service duck type spec](https://k8s-service-bindings.github.io/spec/#provisioned-service) (for the crd status and its secret format)  

**Additional context**

coro:
This PR allows for the creation & deletion of users. Update of users & their credentials is not currently supported.

TODOs:

~- Generate a Secret containing the credentials for the user~
  ~- We can make the user controller own the generated Secret, so that any updates to it will trigger the User controller to reconcile again~
~- Add a Status to the User object, with a `.status.credentials` key pointing to a credentials Secret~
- Allow the provision of the user with a pre-defined password through a Secret supplied in the object Spec
- Allow a User to be updated, either by:
  - Updating the User Spec in the User object
  - Updating the credential Secret generated when the User is first created

This closes #3 


MirahImage:
I'm wondering about the secret structure and how it's likely to be used, including potentially conforming to [provisioned service secret specification](https://k8s-service-bindings.github.io/spec/#provisioned-service) to allow for ducktyping. In particular, I suspect that we would like to have the secret contain both username and password fields (potentially decoupling the resource name from the username).

api/v1alpha1/user_types.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UserSpec defines the desired state of User
+type UserSpec struct {
+	// +kubebuilder:validation:Required
+	Name string    `json:"name"`
+	Tags []UserTag `json:"tags,omitempty"`
```
ChunyiLyu, line 21: if you leave []UserTag empty, what would happen?

coro: You get a user that has signalling access (e.g. AMQP) but no management access:
image

(Corresponds to `(None)` in [this table](https://www.rabbitmq.com/management.html#permissions))

controllers/user_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *UserReconciler) deleteUser(ctx context.Context, client *rabbithole.Client, user *topologyv1alpha1.User) error {
+	logger := ctrl.LoggerFrom(ctx)
+
+	if err := validateResponse(client.DeleteUser(user.Spec.Name)); err != nil {
```
ChunyiLyu, line 207: You can use helper`validateResponseForDeletion()` https://github.com/rabbitmq/messaging-topology-operator/blob/main/controllers/utils.go#L117 here. We shouldn't requeue on already deleted users, same as queues and exchanges.

internal/user_settings.go
```diff
@@ -0,0 +1,55 @@
...
+
+	var userTagStrings []string
+	for _, tag := range tags {
+		userTagStrings = append(userTagStrings, string(tag))
+	}
+
+	return rabbithole.UserSettings{
+		Name:             username,
+		Tags:             strings.Join(userTagStrings, ","),
+		PasswordHash:     rabbithole.Base64EncodedSaltedPasswordHashSHA512(password),
```
ChunyiLyu, line 40: Personally I'd add a comment here to explain why we are using this particular encoding, e.g. what you have written in the test about 'See https://www.rabbitmq.com/passwords.html#computing-password-hash'

internal/user_settings.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"strings"
+
+	rabbithole "github.com/michaelklishin/rabbit-hole/v2"
+	topologyv1alpha1 "github.com/rabbitmq/messaging-topology-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GenerateUserSettings(credentials *corev1.Secret, tags []topologyv1alpha1.UserTag) (rabbithole.UserSettings, error) {
+	username, err := decodeField(credentials, "username")
```
ChunyiLyu, line 23: I would add a comment explaining that keys `username` and `password` are conforming to the provisioned service binding spec, and they are not random keys we have chosen ourselves

controllers/user_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	credentialSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      user.ObjectMeta.Name + "-user-credentials",
+			Namespace: user.ObjectMeta.Namespace,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"username": []byte(base64.StdEncoding.EncodeToString([]byte(user.Spec.Name))),
```
ChunyiLyu, line 122: I would add a comment explaining that keys username and password are conforming to the provisioned service binding spec, and they are not random keys we have chosen ourselves

system_tests/user_system_tests.go
```diff
@@ -0,0 +1,100 @@
...
+		}))
+		Expect(userInfo.PasswordHash).NotTo(BeEmpty())
+
+		By("Creating a Secret with the generated credentials")
+		generatedSecretKey := types.NamespacedName{
+			Name:      "user-test-user-credentials",
+			Namespace: namespace,
+		}
+		var generatedSecret = &corev1.Secret{}
+		Expect(k8sClient.Get(ctx, generatedSecretKey, generatedSecret)).To(Succeed())
```
ChunyiLyu, line 66: You can assert that the created secret keys `username` and `password`. We are conforming to a specific service spec with these keys, and should have a test for them :)

ChunyiLyu (commented):
See comments. Overall looks good

ChunyiLyu:
Question: what would happen if the created credential secret was modified (.data or annotations/labels)? Does that trigger a reconcile() since the controller is watching object? Would you ended up having a newly generated user secret as a result of that?

coro:
> Question: what would happen if the created credential secret was modified (.data or annotations/labels)? Does that trigger a reconcile() since the controller is watching object? Would you ended up having a newly generated user secret as a result of that?

What I'd like to happen is it triggers the reconcile, and is treated as the ultimate source of truth. So a new secret is not created, and the password hash of the user is regenerated and updated if the secret data was changed. 

Having said that, what happens if the data for the username is changed? You then would be left with a User object that is tied to a Secret containing credentials for a different user. Not quite sure how to deal with that just yet.

ChunyiLyu (approved):

PR #34, opened 2021-03-08
Add quick start to README and example manifest for users

ChunyiLyu:

## Summary Of Changes

- Add quick start and links to example manifests to README
- Add an example to create user

### DIFF
--- a/docs/examples/users/user.yaml
+++ b/docs/examples/users/user.yaml
@@ -1,0 +1,11 @@
+apiVersion: rabbitmq.com/v1alpha1
+kind: User
+metadata:
+  name: user-sample
+spec:
+  name: test-user # name of the user
+  tags:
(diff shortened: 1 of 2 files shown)


docs/examples/users/user.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rabbitmq.com/v1alpha1
+kind: User
+metadata:
+  name: user-sample
+spec:
+  name: test-user # name of the user
+  tags:
+  - management # available tags are 'management', 'policymaker', 'monitoring' and 'administrator'
```
coro, line 8: There needs to be a rabbitmqClusterReference in this too.

coro (changes requested):

docs/examples/users/user.yaml
ChunyiLyu: Updated, thanks for catching it!

coro (approved):

Issue #35, opened 2021-03-08
Support updates to User resources & credential Secrets


coro:
**Is your feature request related to a problem? Please describe.**
Currently, User objects can only be created and deleted, but not updated. 

**Describe the solution you'd like**
### Editing the User
- Changing User.Spec.Name or User.Spec.RabbitmqClusterReference: TBD, but probably reject the request with a validating webhook, rather than delete the existing user and create a new one. Otherwise, an app consuming the User Secret will have to reload the secret to regain access to an authenticated user.
- Changing User.Spec.Tags: Updates the tags associated with a user.

### Editing the User Secret
- Changing Secret.Data["username"]: TBD, but again probably reject the request
- Changing Secret.Data["password"]: Calculate a new password hash and update the user

**Additional context**
Original Issue: #3

carlosjgp:
could we reopen this?

cronventis:
Please implement :)

PR #44, opened 2021-03-08
Allow pre-definition of user passwords

### LINKED ISSUE #36: Allow pre-defined passwords to be used for created User objects
As a user, I would like to optionally provide my own pre-defined passwords for generated User objects rather than rely on those generated by the operator.

Allow for a new field `User.Spec.importPasswordSecret` which is a `LocalObjectReference` to a Secret containing the password to use when declaring the User. This Secret must contain a key whose value will become the password of the new user:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: credentialsSecret
  namespace: testing
type: Opaque
data:
  credentialList.importUser.password: d2h5YXJleW91bG9va2luZ2hlcmU=
---
apiVersion: rabbitmq.com/v1alpha1
kind: User
metadata:
  name: import-user
  namespace: testing
spec:
  name: import-user
  rabbitmqClusterReference:
    name: rmq
    namespace: rabbitmq-system
  importPasswordSecret:
    name: credentialsSecret
    passwordKey: credentialList.importUser.password # defaults to just 'password'
```

**Handling Updates:**
The initial Secret is not the 'source of truth' for the password, but only the initial source. Once the credential Secret is generated for a user, that is considered the source of truth and any further updates to, or deletion of, the initial secret are ignored.

**Additional context**
Original Issue: #3

coro:
This PR allows a user to specify a Secret which contains a password to set on the User. The controller looks for the username & password under the keys `username` & `password` in the Secret. This Secret must be in the same namespace as the User, in order to mitigate privilege escalation.

Note that this import only happens upon creation - once the User is created, the import Secret is ignored. A user must be deleted & recreated in order to change a password or username until #35 is played.

This closes #36 

## Summary Of Changes
This PR adds several new fields to allow import of passwords from a created Secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: credentials-secret
  namespace: testing
type: Opaque
data:
  username: bXktc2VjcmV0LXVzZXI=
  password: d2h5YXJleW91bG9va2luZ2hlcmU=
---
apiVersion: rabbitmq.com/v1alpha1
kind: User
metadata:
  name: import-user
  namespace: testing
spec:
  rabbitmqClusterReference:
    name: rmq
    namespace: rabbitmq-system
  importCredentialsSecret:
    name: credentials-secret
```


coro:
cc @n3wscott as I believe this was one of the knative requirements.

ChunyiLyu (commented):
What we have discussed:
1. Import password and username from secret
2. Delete user.spec.name; if name is not present and no import secret provided, randomly generates username and password pair
3. two example manifests about how to create users: randomly generated username/password and pre defined password and username in `docs/examples/users`

Zerpet (approved):

ChunyiLyu (approved):
LGTM. There are several places in the controller which a failure was only returned but not logged or recorded as event. But I want to approve the PR so we can merge it soon and avoid merge conflicts. Logging and publish events for all errors can come in a different PR.

PR #52, opened 2021-03-12
Allow topology objects to be deleted if cluster deleted

coro:
Prior to this PR, if a RabbitmqCluster was deleted prior to a topology object referencing that cluster, any attempts to delete the topology object would fail, due to the finalizer step failing. This was because at the start of every Reconcile, the controller attempts to create a rabbit-hole client to the cluster, which would fail.

## Summary Of Changes

The function which retrieves connection info about the RabbitmqCluster now returns a `NoSuchRabbitmqClusterError`, which can be type checked in the Reconcile function, provided the error is wrapped correctly. Any controller which attempts to reconcile a deletion of an object which gets this error can continue with removing the finalizer and ignoring the error creating a rabbit-hole client (the cluster was deleted, so no longer requires the object to be deleted as well).

## Additional Context

This problem was really getting in the way of my development workflow - if a system test failed, I would have to manually edit the objects to remove their finalizers rather than just `k delete`ing them.


system_tests/deletion_system_tests.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+		}
+		Expect(k8sClient.Create(ctx, &exchange)).To(Succeed())
+		Expect(k8sClient.Create(ctx, &policy)).To(Succeed())
+		Expect(k8sClient.Create(ctx, &queue)).To(Succeed())
+		Expect(k8sClient.Create(ctx, &user)).To(Succeed())
+		Expect(k8sClient.Create(ctx, &vhost)).To(Succeed())
+	})
+
+	FIt("handles the referenced RabbitmqCluster being deleted", func() {
```
coro, line 92: !

controllers/queue_controller.go
```diff
@@ -127,6 +135,11 @@ func (r *QueueReconciler) addFinalizerIfNeeded(ctx context.Context, q *topologyv
 func (r *QueueReconciler) deleteQueue(ctx context.Context, client *rabbithole.Client, q *topologyv1alpha1.Queue) error {
 	logger := ctrl.LoggerFrom(ctx)
 
+	if client == nil {
+		logger.Info("Not deleting queue as RabbitmqCluster is already deleted or down", "queue", q.Name)
```
ChunyiLyu, line 139: I thought we are not deleting queues when the RabbitmqCluster is deleted/gone (when there is no secret object to generate a rabbithole client). What do you mean by the word `down` in sentence `RabbitmqCluster is already deleted or down`?

system_tests/utils.go
```diff
@@ -143,3 +145,36 @@ func kubernetesNodeIp(ctx context.Context, clientSet *kubernetes.Clientset) stri
 	}
 	return nodeIp
 }
+
+func setupTestRabbitmqCluster(name, namespace string) *rabbitmqv1beta1.RabbitmqCluster {
```
ChunyiLyu, line 149: Nice!

controllers/queue_controller.go
coro: This was down to my uncertainty about what 
```go
import client "sigs.k8s.io/controller-runtime/pkg/client"

client.Client{}.Get(ctx, types.NamespacedName{Name: rmq.Name, Namespace: rmq.Namespace}, cluster)
```
returning an error would signify. The Get returns an error if the cluster has been deleted, but are there other occasions where it would fail to find the object in the cache? Perhaps during a cluster restart? In which case, the object would be down but not deleted. I don't know if they can be distinguished between (but happy to be corrected!)

coro (commented):
LGTM, @ChunyiLyu

Issue #43, opened 2021-03-08
All resources have a minimum status


n3wscott:
**Is your feature request related to a problem? Please describe.**

A controller or operator needs feedback to know if the controller accepted and realized the spec provided.

**Describe the solution you'd like**
All resources, at minimum have the following status shape:

```yaml
status:
  conditions:
  - lastTransitionTime: "2021-02-23T20:52:21Z"
    status: "True"
    type: Ready
  observedGeneration: 1
```

**Describe alternatives you've considered**

Please do this 😉

michaelklishin:
@ChunyiLyu will this rely on the Kubernetes resource state or query RabbitMQ? The latter can be pretty expensive with, say, 100s of thousands of queues.

ChunyiLyu:
@michaelklishin My plan is to mark resources as 'ready' after successful Create/Update calls to the server. As you said, query rabbit is expensive and I'm not planning on doing that.

PR #56, opened 2021-03-12
Add and set Status Conditions 'Ready' in all resources

ChunyiLyu:

## Summary Of Changes

- Per feedback from https://github.com/rabbitmq/messaging-topology-operator/issues/43, add a status condition to indicate that the CR was successfully created or updated by the controllers.
- Status condition 'Ready'  looks like:
```
  conditions:
  - lastTransitionTime: "2021-02-23T20:52:21Z"
    status: "True"
    type: Ready
    reason: SuccessfulCreateOrUpdate
    message: ""
```
- 'Ready' status condition is set for all resources: queues, vhost, users, vhost, policies, exchanges, and bindings
- Extracted common error messages from controllers and created "controllers/common.go"


coro (approved):
Looks great, and I really like the error & conditions extraction.

This has made me think even more about the fact that a lot of our resources are rather similar. In broad strokes, create/update operations involve a declare through rabbithole, and delete operations require a delete. They all have a method to convert from k8s API JSON to rabbithole.XXXSettings{}. A lot of the controller reconcile code is copied between the objects.

I've been wondering whether this is a sign that we should have a common reconcile function, and use interfaces to provide the object-specific behaviours (which aren't very many). But possibly a refactor for another time.

PR #65, opened 2021-03-15
Support Bindings deletion

### LINKED ISSUE #64: Delete bindings.rabbitmq.com

**Additional context**

ChunyiLyu:
This closes #64 

## Summary Of Changes

- Support `bindings.rabbitmq.com deletion` when no bindings arguments are provided
- Bindings properties key is needed to delete binding; help function `GeneratePropertiesKey()` is used. Properties keys are generated by rabbitmq server when bindings are created:
1) when routing key and argument are not provided, properties key is `~`.
2) when routing key is set and no argument is provided, properties key is the routing key itself
3) if routing key has character `~`, it's replaced by `%7E`.
4) when arguments are provided, properties key is the routing key (could be empty) plus the hash of arguments. The hash function used is erlang specific, and implementing the escape logic is not straightforward. I've decided to not support deletion of bindings if arguments are provided for now.
- See: https://github.com/rabbitmq/rabbitmq-server/blob/52479099ecc9807097ad0af70b6c5d54cb5031e7/deps/rabbitmq_management_agent/src/rabbit_mgmt_format.erl#L297 for how the server is generating binding properties key

## Follow up

We should support deletion for bindings with arguments because we allow users to configure `bindings.spec.arguments`. We can list the bindings between queue/exchanges or exchange/exchange to find the binding properties key first, then delete. Reimplement the function which generates properties key for bindings with arguments could be complicated. Follow up issue: https://github.com/rabbitmq/messaging-topology-operator/issues/66

Thanks @mkuratczyk for reading erlang ;)


ChunyiLyu:
Closing in favor of this issue: https://github.com/rabbitmq/messaging-topology-operator/issues/66

PR #74, opened 2021-03-18
validation webhook for queues.rabbitmq.com and exchanges.rabbitmq.com

ChunyiLyu:

## Summary Of Changes

- validation webhook for `queues.rabbitmq.com` and `exchanges.rabbitmq.com`
- validation webhook only checks for updates in the future. we should consider input validation for create in the future as well. 
- for queues, you get: 1) error type 'forbidden' for updates that the controller chooses to disallow: queue name/vhost/rabbitmqClusterReference; 2) error type 'invalid' for updates that will be rejected by rabbitmq server: queue types/autoDelete/durable. I chose to ignore queue.spec.arguments for now, since we might want to create policies under the hood in the future.
- for exchanges, you get: 1) error type 'forbidden' for updates that the controller chooses to disallow: queue name/vhost/rabbitmqClusterReference; 2) error type 'invalid' for updates that will be rejected by rabbitmq server: exchange types/autoDelete/durable; updates on exchanges arguments are accepted by the server

## Additional Context

Test coverage in unit and system tests.

### DIFF
--- a/api/v1alpha1/exchange_types.go
+++ b/api/v1alpha1/exchange_types.go
@@ -12,6 +12,7 @@ package v1alpha1
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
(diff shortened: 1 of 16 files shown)


MirahImage (approved):
Tried it out, works as expected

PR #87, opened 2021-03-26
Generate clients

vaikas:
This closes #

## Summary Of Changes
Add codegen generated clients so that these can be used from things like github.com/knative-sandbox/eventing-rabbitmq. While it seems like a lot of new code, it's all code generated. Also had to fix couple of tags / add methods in the api/v1alpha2 so the code-gen behaves properly.

Start of the usage can be seen here (by no means complete, just starting that work over there):

https://github.com/knative-sandbox/eventing-rabbitmq/pull/267


hack/boilerplate.go.txt
```diff
@@ -0,0 +1,6 @@
+/*
```
MirahImage, line 1: Does it make sense to use a separate file here rather than re-use NOTICE.go.txt

MirahImage (commented):
Does the script `hack/update-codegen.sh` need to be added to the `make generate` target?

n3wscott:
> Does the script `hack/update-codegen.sh` need to be added to the `make generate` target?

Does kubebuilder now encourage a Makefile? This has been seen as an anti-pattern in many projects, including Knative.

hack/boilerplate.go.txt
vaikas: Yep, totes missed the NOTICE.go.txt, updated.

MirahImage (approved):

PR #99, opened 2021-04-07
Use the correct status reference and stop generating client set all the time

ChunyiLyu:

## Summary Of Changes

- separate client set generation to its own make target and remove `go mod vendor` from installing tools
- add a line to remind contributors to re generate client set when create PRs
- use cluster.status.binding to get secret name instead of cluster.status.defaultUser.secretReference

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ install-tools:
 	# This one just needs to be fetched and not installed, get & mod so it ends up in the right place.
 	# Note we grep it out above, and just do a go get & go mod for it.
 	go get -d k8s.io/code-generator
-	go mod vendor
 
 unit-tests: install-tools generate fmt vet manifests ## Run unit tests
 	ginkgo -r --randomizeAllSpecs api/ internal/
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -85,6 +84,12 @@ vet:
 # Generate code & docs
 generate: install-tools api-reference
 	controller-gen object:headerFile="hack/NOTICE.go.txt" paths="./..."
+
+generate-client-set:
+	# This one just needs to be fetched and not installed, get & mod so it ends up in the right place.
+	# Note we grep it out above, and just do a go get & go mod for it.
+	go get -d k8s.io/code-generator
+	go mod vendor
```
coro, line 92: Are we sticking with vendoring then?

coro (approved):
LGTM. One question, but shouldn't block you merging.

Makefile
ChunyiLyu: Only when you are generating the code. The code generation script `./hack/update-codegen.sh` refers to ` ./vendor/k8s.io/code-generator` at the moment. After running the the make target, `vendor/` can be safely deleted and you don't need to check it in.

coro: I feel a bit weird having a package-level structure (i.e. `vendor/`) existing just for this one script, even if it's temporary for this one script.

Having said that, it's in the gitignore and you didn't introduce it in this PR, so feel free to go ahead with this change regardless.

PR #113, opened 2021-04-15
Support for setting schema replication parameters

### LINKED ISSUE #112: Add support for setting global parameter `schema_definition_sync_upstream`
**Is your feature request related to a problem? **
Schema replication upstream endpoints information is configured with global parameters. We want to have a method to configure this via the Kubernetes API. It's currently set through cli command like `rabbitmqctl set_schema_replication_upstream_endpoints '{"endpoints": ["a.rabbitmq.eu-1.local:5672","b.rabbitmq.eu-1.local:5672","c.rabbitmq.eu-1.local:5672"], "username": "schema-replicator", "password": "$3kRe7"}'`

Describe the solution you'd like

Instead of providing an API for people to set any global parameters, having a dedicated `SchemaReplication` crd suits the use case and allows operator to perform better input validation. Global parameters also contains information like cluster name and internal cluster ID, which are not values we would like users to manipulate. The crd will take a secret that contains keys `username`, `password`, and `endpoints`. `username` and `password` are credentials that shouldn't be stored in manifests. 

Describe alternatives you've considered

Additional context

ChunyiLyu:
This closes #112 

## Summary Of Changes

- add a crd `schemareplications.rabbitmq.com` crd which takes in `spec.upstreamSecret` and `spec.rabbitmqClusterReference`
- webhook which does not allow updates on `spec.rabbitmqClusterReference`
- the secret that  `spec.upstreamSecret`  refers to must have keys `username`, `password`, and `endpoints`. When multiple endpoints are provided, they need to be separated by ','


config/samples/rabbitmq.com_v1alpha2_schemareplication.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: rabbitmq.com/v1alpha2
+kind: SchemaReplication
+metadata:
+  name: schemareplication-sample
+spec:
+  # Add fields here
+  foo: bar
```
MirahImage, line 7: Did we mean to leave this as the sample YAML

controllers/suite_test.go
```diff
@@ -133,6 +142,21 @@ var _ = BeforeSuite(func(done Done) {
...
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "endpoints-secret",
+			Namespace: "default",
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"username":  []byte("a-random-user"),
+			"password":  []byte("a-random-password"),
+			"endpoints": []byte("a.endpoints.local:5672,b.endpoints.local:5672,c.endpoints.local:5672"),
```
MirahImage, line 155: Do we want to leave this as a single comma-separated text field, or would it make more sense for it to be an array?

ChunyiLyu: It's a secret object. I don't think you can have anything other than string as the value.

config/samples/rabbitmq.com_v1alpha2_schemareplication.yaml
ChunyiLyu: Actual example lives in `docs/example`. Manifests `config/samples` are generated by kubebuilder and not actually valid. I can create a separate PR to get rid of all of them.

PR #114, opened 2021-04-15
Example manifests are in docs/examples

ChunyiLyu:
This closes #

## Summary Of Changes

Remove `config/samples/` because example manifests are in docs/examples.  Manifests in `config/samples/` are not maintained and outdated.


MirahImage (approved):
Awesome! I had never noticed these samples before, but was very confused by it while reviewing the previous PR

PR #134, opened 2021-05-14
Add CRD descriptions on fields that cannot be updated

ChunyiLyu:

## Summary Of Changes

As part of the effort of adding documentation for topology operator in rabbitmq.com (https://github.com/rabbitmq/rabbitmq-website/pull/1182), I realized that we are not documenting which fields cannot be updated in any CRDs api description. This PR adds description to all CRDs fields that cannot be updated.

### DIFF
--- a/api/v1beta1/binding_types.go
+++ b/api/v1beta1/binding_types.go
@@ -17,18 +17,23 @@ import (
 
 // BindingSpec defines the desired state of Binding
 type BindingSpec struct {
-	// Default to vhost '/'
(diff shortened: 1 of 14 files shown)


coro (approved):
LGTM. I think we should update these pre-1.0 to have more descriptive info but this is good for now

ChunyiLyu:
> LGTM. I think we should update these pre-1.0 to have more descriptive info but this is good for now

The API docs are rather concise currently 😄

Issue #126, opened 2021-05-06
The Operator could watch objects it depends on to reconcile


Zerpet:
Some kinds e.g. User depend on a user provided Secret. CRUD operations to this secret won't trigger a reconcile loop, which in turn results in the Secret containing information that doesn't match RabbitMQ's internal database. To "force a sync", a user has to modify the kind in question, for example, by adding a label.

The Operator could watch those external objects and trigger a reconcile on update of external resources.

ChunyiLyu:
This enhancement applies to `federations.rabbitmq.com`.

Zerpet:
I'd like to see this enhancement into all kinds that reference a Secret in their spec. This caught me a few times while working with `schemareplication` kind, but I can see benefit of adding this to all other kinds.

PR #144, opened 2021-05-27
Refactor federation spec.uri to spec.uriSecret

ChunyiLyu:
This closes #

## Summary Of Changes

- federation upstream uri contains credentials and should not be saved in a CR manifest directly
- controller does not watch the uriSecret for now; this will be addressed in a separate issue and PR that will add same function to schema replication CR as well (issue: #126).


config/crd/bases/rabbitmq.com_federations.yaml
```diff
@@ -74,17 +74,24 @@ spec:
...
-                description: The AMQP URI(s) for the upstream. Required property.
-                type: string
+              uriSecret:
+                description: Secret contains the AMQP URI(s) for the upstream. The
+                  Secret must contain the key `uri` or operator will error. Required
+                  property.
+                properties:
+                  name:
+                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      TODO: Add other useful fields. apiVersion, kind, uid?'
```
MirahImage, line 84: Is the `TODO` still meant to be here?

ChunyiLyu: It comes from kubernetes api corev1.LocalObjectReference directly: https://github.com/kubernetes/api/blob/b0d9a0e45e6f6f7c5f18d76d950a4d19e41d1f50/core/v1/types.go#L5407

system_tests/federation_system_test.go
```diff
@@ -51,7 +64,7 @@ var _ = Describe("federation", func() {
 
 		Expect(upstream.Name).To(Equal(federation.Spec.Name))
 		Expect(upstream.Vhost).To(Equal(federation.Spec.Vhost))
-		Expect(upstream.Definition.Uri).To(Equal(federation.Spec.Uri))
+		Expect(upstream.Definition.Uri).To(Equal(federationUri))
```
coro, line 67: I'm surprised this passes, given one is a string and the other a `corev1.Secret`. Am I missing something?

coro: Ah I see, it's a naming collision with the string in line 21, being redefined in line 25. Could you rename the secret please?

coro (changes requested):

system_tests/federation_system_test.go
ChunyiLyu: Updated the name of the secret.

PR #177, opened 2021-07-01
Able to reference a User from permissions.rabbitmq.com crd

### LINKED ISSUE #165: Associate permissions to autogenerated users
When creating `User` objects with autogenerated credentials (so without using `importCredentialsSecret`), we cannot create `Permission` objects for them because the username is a randomly generated string.

A good solution would be to not autogenerate the username, and instead use the `User` `metadata.name` field when creating the RabbitMQ user and populating the credentials secret.

Another option is to allow referencing a `User` object in the `PermissionSpec`, so that the operator can retrieve the generated username from the secret and use it in the permission.

Example:
```yaml
kind: User
metadata:
  name: example
spec:
  rabbitmqClusterReference:
    name: example-broker
---
kind: Permission
metadata:
  name: example-user
spec:
  userRef:
    name: example # this is the User above
  vhost: '/'
  permissions:
    configure: '.*'
    write: '.*'
    read: '.*'
  rabbitmqClusterReference:
    name: example-broker

```

**Additional context**
Without this feature, when using autogenerated credentials the operator fails to create `Permission` objects with the error `failed to set permission    {"reconciler group": "rabbitmq.com", "reconciler kind": "Permission", "name": "example-user", "namespace": "example", "user": "example", "vhost": "/", "error": "Error 400 (bad_request): vhost_or_user_not_found"}`

ChunyiLyu:
This closes #165 

## Summary Of Changes

- add a field `spec.userReference` (type LocalObjectReference) to permissions.rabbitmq.com
- preserve `spec.user` to minimize disruption to our users
- `spec.user` is no longer a mandatory field. However, permissions webhook will make sure that one of  `spec.userReference` and `spec.user` must be provided; webhook also returns an invalid error if both fields are specified.
- tested in system tests
- a small refactor in integration tests to use `SatisfyAll` where it makes sense

### Why reference user in `permissions.rabbitmq.com` but not the other around?

To assign permissions for a users, we need to have the username, permissions configurations, and the vhost. Users are not vhost specific and can have different permissions configurations in different vhosts. If we were to add permissions configurations in `users.rabbitmq.com`, updates that revokes permissions in certain vhosts could be hard to perform (operator will need to keep track of the previous list of vhosts and check if there were deletion on every updates).


MirahImage (approved):
LGTM, but integration tests are failing

PR #197, opened 2021-07-27
Remove validation on exchange.spec.type

### LINKED ISSUE #192: Exchanges definition - Allow other types from plugins such as rabbitmq-consistent-hash-exchange
We are currently migration from our old RabbitMQ chart to the Operator, however, we are not able to migrate some of the exchange as they use the following plugin: https://github.com/rabbitmq/rabbitmq-consistent-hash-exchange
We are getting the following error: 
> The Exchange "rabbitmq-cluster-event" is invalid: spec.type: Unsupported value: "x-consistent-hash": supported values: "direct", "fanout", "headers", "topic"

If I check in the CRD definition I see that there is a list that is mentioned as enum. I assume this is done to ensure the validation.
```              
type:
 default: direct
 description: Cannot be updated
 enum:
     - direct
     - fanout
     - headers
     - topic
```
We would like to have the option to declare these exchange directly via the operator. Considering the plugin is from the rabbitmq github project, we would like to have that possibility.
We tried to alter the manifest of the CRD to add that type of exchange in the enum list. And it seems to work, the exchange is properly creatd. However, we will have to patch it again after every update of the CRD. If a more permanent solution could be provided, this would be better for us. And this could affect other plugins (that we aren't using)
The alternate solution that we have is to create these exchange outside of the CRD via K8S job as we did in the past, but this defeat a bit the purpose of switching to the CRD for us.

**Additional context**
N/A

ChunyiLyu:
This closes #192 

## Summary Of Changes

Other types of exchanges could be available in a cluster by enabling plugins like consistent hash or sharding. Removing the type validation on `exchange.spec.type` allows additional exchange types to be used.

### DIFF
--- a/api/v1beta1/exchange_types.go
+++ b/api/v1beta1/exchange_types.go
@@ -24,7 +24,6 @@ type ExchangeSpec struct {
 	// +kubebuilder:default:=/
 	Vhost string `json:"vhost,omitempty"`
 	// Cannot be updated
-	// +kubebuilder:validation:Enum=direct;fanout;headers;topic
 	// +kubebuilder:default:=direct
 	Type string `json:"type,omitempty"`
 	// Cannot be updated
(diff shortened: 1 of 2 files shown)


Zerpet (approved):
:shipit:

PR #203, opened 2021-07-30
Finalizer refactor

ChunyiLyu:

## Summary Of Changes

- extract universal helpers to remove and add finalizers used for all controllers because: 1) deletion finalizers are all in the same format 2)logic to add and remove finalizers is the same across controllers
- added finalizers tests to make sure added finalizers hasn't changed (has to stay the same for compatibility)
- updated controller-gen annotations in all CRD

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -10,10 +10,14 @@ This product may include a number of subcomponents with separate copyright notic
 package controllers
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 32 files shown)


controllers/utils.go
```diff
@@ -55,3 +59,36 @@ func serviceDNSAddress(svc *corev1.Service) string {
...
+	controllerutil.RemoveFinalizer(obj, finalizer)
+	if err := client.Update(ctx, obj); err != nil {
+		return fmt.Errorf("failed to delete finalizer: %w", err)
+	}
+	return nil
+}
+
+// deletionFinalizer returns generated deletion finalizer
+// finalizers follow the format of deletion.finalizers.kind-plural-form.rabbitmq.com
+// for example: deletion.finalizers.bindings.rabbitmq.com and deletion.finalizers.policies.rabbitmq.com
```
coro, line 85: Could these just all be the same finalizer name? Would that be a problem?

coro (approved):
1 question, otherwise ship it!

controllers/utils.go
ChunyiLyu: For a new operator, yes there won't be a problem. However for topology operator which already has previous versions, we need to make sure that finalizers stay the same to maintain compatibility with previous versions.

PR #221, opened 2021-08-24
Fix flaky finalizer test

Zerpet:

## Summary Of Changes
Flakes are great in corn 🌽 but not in your tests.

## Additional Context
Chose to declare the error in a var and then check if it's nil, rather than declaring everying inside the `if` condition, for better readability.


ChunyiLyu (approved):
Good one!

PR #240, opened 2021-09-22
Add spec.endpoints for schemareplication.rabbitmq.com

ChunyiLyu:
This closes #

## Summary Of Changes

Feedback and suggestion from @pjk25 🥳 

Add spec.endpoints for schemareplication.rabbitmq.com, so that: 
 - users can then reference the default user secret
 - it might be better for vault integration in the future because username and password are fetched from vault, but not the list of endpoints. So this creates better separation
 - this change is not a breaking change, because we can still let people specify list of endpoints in the secret, but the top level endpoints property takes precedence

## Additional Context

Tested in unit and system tests.


pjk25:
@ChunyiLyu changes look good to me from a visual inspection, though I do not understand the failures in the system tests.

pjk25 (approved):

PR #257, opened 2021-11-10
Downgrade code-generator to 0.21.x

gvmw:
This was introduced part of https://github.com/rabbitmq/messaging-topology-operator/pull/87, and in the latest eventing-rabbitmq changes it became obvious that the versions started to drift. This discussion explains the problem best: https://github.com/rabbitmq/messaging-topology-operator/discussions/256

## Summary Of Changes

- Downgrade code-generator to v0.21.4
- Upgrade all other related components to the same version

## Additional Context

https://github.com/knative-sandbox/eventing-rabbitmq/pull/492#issuecomment-963276439

This change lines up all versions, and the local unit & integration tests pass. My next step is to check if CI remains green, and if the change actually works in the context of eventing-rabbitmq, which is how this started.


Zerpet (approved):
Looks good to me 👍

ChunyiLyu (approved):

gvmw:
Thank you for the pro-active approach, I was just going to ask for a merge 🙂

When are you thinking of cutting a new messaging-topology-operator version? There is no rush, I will be pinning the go dep to a specific sha until then.

These are the important dates from Knative release schedule perspective: https://github.com/knative/community/blob/9b9e96773cfc04ca05220939e37df19a45a11852/mechanics/RELEASE-SCHEDULE.md

PR #281, opened 2022-01-10
Add SuperStreamCRD

coro:
This PR is the successor of #263 , including only the SuperStreamCRD. The more experimental SuperStreamConsumer CRD will be included in a separate controller once this PR is merged, so that the topology operator does not end up with permissions to manage Pods.


PROJECT
```diff
@@ -94,4 +94,22 @@ resources:
...
+  version: v1beta1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: rabbitmq.com
+  group: rabbitmq.com
+  kind: SuperStreamConsumer
+  path: github.com/rabbitmq/messaging-topology-operator/api/v1beta1
+  version: v1beta1
```
Zerpet, line 114: I'd prefer to have the version as alpha, so that we don't tie ourselves into maintaining these CRDs for many releases afterwards. IIUC, some of the functionality here will be build into RabbitMQ Core and the intention is to fill an existing gap. It shounds to me as an alpha (experimental) state, more than a beta (pre-release stable).

config/default/webhookcainjection_patch.yaml
```diff
@@ -0,0 +1,15 @@
+# This patch add annotation to admission webhook config and
+# the variables $(CERTIFICATE_NAMESPACE) and $(CERTIFICATE_NAME) will be substituted by kustomize.
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
Zerpet, line 8: This whole [file seems duplicated](https://github.com/rabbitmq/messaging-topology-operator/blob/super-streams-only/config/default/base/webhookcainjection_patch.yaml). The mutating webhook bit makes me feel a bit uneasy. We could remove the file as a whole, or remove the mutating webhook bit at least?

config/webhook/manifests.yaml
```diff
@@ -206,3 +206,23 @@ webhooks:
...
+      name: webhook-service
+      namespace: system
+      path: /validate-rabbitmq-com-v1beta1-superstream
+  failurePolicy: Fail
+  name: vsuperstream.kb.io
+  rules:
+  - apiGroups:
+    - rabbitmq.com
+    apiVersions:
+    - v1beta1
```
Zerpet, line 222: Shouldn't this be `v1alpha1`?

controllers/super_stream_controller.go
```diff
@@ -0,0 +1,183 @@
...
+			"SuperStreams cannot be scaled down: an attempt was made to scale from %d partitions to %d",
+			len(superStream.Status.Partitions),
+			superStream.Spec.Partitions,
+		)
+		msg := fmt.Sprintf("SuperStream %s failed to reconcile", superStream.Name)
+		logger.Error(err, msg)
+		if writerErr := r.SetReconcileSuccess(ctx, superStream, topology.NotReady(msg, superStream.Status.Conditions)); writerErr != nil {
+			logger.Error(writerErr, failedStatusUpdate, "status", superStream.Status)
+		}
+		return reconcile.Result{}, err
```
Zerpet, line 76: Returning the error is going to trigger another recocile loop, which will never succeed unless the user takes action and changes the partitions.

I would be inclined to return `Result{}, nil` instead, perhaps produce an event, on top the setting the reconcile success condition to false.

controllers/super_stream_controller_test.go
```diff
@@ -0,0 +1,678 @@
...
+							"Type":    Equal("direct"),
+							"Durable": BeTrue(),
+							"RabbitmqClusterReference": MatchAllFields(Fields{
+								"Name":      Equal("example-rabbit"),
+								"Namespace": Equal("default"),
+							}),
+						}))
+
+					})
+					By("creating n stream queue partitions", func() {
```
Zerpet, line 94: nit: empty line in between, for consistency

controllers/super_stream_controller_test.go
```diff
@@ -0,0 +1,678 @@
...
+								})),
+								"RoutingKey": Equal(strconv.Itoa(i)),
+								"RabbitmqClusterReference": MatchAllFields(Fields{
+									"Name":      Equal("example-rabbit"),
+									"Namespace": Equal("default"),
+								}),
+							}))
+						}
+					})
+					By("setting the status condition 'Ready' to 'true' ", func() {
```
Zerpet, line 159: nit: empty line because _consistency_

controllers/super_stream_controller_test.go
```diff
@@ -0,0 +1,678 @@
...
+						)
+
+						return superStream.Status.Conditions
+					}, 10*time.Second, 1*time.Second).Should(ContainElement(MatchFields(IgnoreExtras, Fields{
+						"Type":   Equal(topology.ConditionType("Ready")),
+						"Reason": Equal("SuccessfulCreateOrUpdate"),
+						"Status": Equal(corev1.ConditionTrue),
+					})))
+				})
+				When("a binding is deleted", func() {
```
Zerpet, line 194: nit: you know by now my nit picks

controllers/super_stream_controller_test.go
```diff
@@ -0,0 +1,678 @@
...
+								)
+								if err != nil {
+									return false
+								}
+								return binding.CreationTimestamp != initialCreationTimestamp
+							}, 10*time.Second, 1*time.Second).Should(BeTrue())
+						})
+					})
+				})
+				When("a queue is deleted", func() {
```
Zerpet, line 240: nit: last one 🙈 I'd say either all should have an empty line, or none.

system_tests/superstream_system_test.go
```diff
@@ -0,0 +1,203 @@
...
+				partitionOrder := int(fetchedBinding.Arguments["x-stream-partition-order"].(float64))
+				foundPartitionOrderStreams[partitionOrder] = fetchedBinding
+			}
+			for i := 0; i < superStream.Spec.Partitions; i++ {
+				_, ok := foundPartitionOrderStreams[i]
+				Expect(ok).To(
+					BeTrue(),
+					fmt.Sprintf("Expected to find a partition assigned to all indices, but %d was missing: %+v", i, foundPartitionOrderStreams),
+				)
+			}
```
Zerpet, line 151: We have the same test at integration level. I'm not sure if this test at system level is adding more coverage. WDYT?

Zerpet (commented):
Feel free to ignore all my nit picks 🙈

config/default/webhookcainjection_patch.yaml
coro: Yeh, looks that way. I think kubebuilder added it in when I created a new api with the CLI. It's not referenced by any kustomize file, so I'll remove this one.

system_tests/superstream_system_test.go
coro: Agreed, it made more sense when there was the other CRD but it's less useful now.

Zerpet (approved):

ChunyiLyu (approved):
🚀

Issue #233, opened 2021-09-13
Cannot use operator with tls enabled rabbitmq cluster


olfway:
## Describe the bug
My rabbitmq cluster has tls certificate for *.some.domain.com,
but topology operator tries to connect to https://operator.test-2.svc:15671 and fail with an error:
x509: certificate is valid for *.test-2.some.domain.com, test-2.some.domain.com, not operator.test-2.svc

But for my applications I have to have a valid *.some.domain.com cert and cannot use self-signed one for operator.test-2.svc

There probably should be a way to provide a hostname for topology operator to connect to

## Version and environment information
- Messaging Topology Operator: 1.1.0
- RabbitMQ Cluster Operator: 1.8.3

starkers:
👍🏼  we're experiencing similar frustrations ourselves

Zerpet:
Hey folks, we discussed this issue in our internal sync up and we are going to prioritise it. We think the Operator should work in any TLS setup and this can be considered a bug. Thank you for reporting this!

coro:
This can be an extension of the `RabbitmqClusterReference` object, removing the implicit assumption that the cluster-local address is fine:

```
type RabbitmqClusterReference struct {
	// The name of the RabbitMQ cluster to reference.
	// +kubebuilder:validation:Required
	Name string `json:"name"`
	// The namespace of the RabbitMQ cluster to reference.
	// Defaults to the namespace of the requested resource if omitted.
	// +kubebuilder:validation:Optional
	Namespace string `json:"namespace"`
	// The hostname used to connect to the RabbitMQ Cluster on which this topology object will be created.
        // If unset, defaults to the cluster-local DNS address `name.namespace.svc`
	// +kubebuilder:validation:Optional
	Hostname string `json:"hostname"`
}
```

masterphenix:
Hello, we are currently blocked with this issue also.
We have production rabbitmq clusters on VMs, and are considering migrating them to Kubernetes by using the cluster and messaging topology operators, which are really promising, easy to use, maintain and monitor. 
However, not being able to use a trusted certificate alongside with the Messaging Topology Operator is blocking us, and prevents us from going further at this point.
Any estimate on when a fix could be released ? Or maybe a workaround in the mean time ?

logineimer:
Same here. It'd be great if there was a fix soon.

Zerpet:
[...]

PR #282, opened 2022-01-11
Initial proposal to address #233

Zerpet:
Design document and proposal to address the scenarios discussed in #233. Any and all feedback is welcome.


docs/proposals/20220105-start-certificate-support.md
```diff
@@ -0,0 +1,269 @@
...
+```gherkin
+Background:
+  Given that RabbitMQ is deployed with the Cluster Operator
+  And RabbitMQ only uses protocols over TLS
+  And the certificate presented by RabbitMQ "rmq-cert" is signed by a trusted CA
+  And RabbitMQ is accessible within Kubernetes via `myrabbit.default.svc.cluster.local`
+  And RabbitMQ is accessible externally via Ingress with FQDN `myrabbit.example.com`
+
+Scenario: RabbitMQ presents a star certificate for K8s cluster domain
+  Given the certificate "rmq-cert" is valid for '*.default.svc.cluster.local'
```
coro, line 130: I think these domains should be your custom ones, right?

coro (approved):
LGTM as a proposal. I think this ties us to only supporting clusters with wildcard certificates, but this is a good first step.

One question I have; if the external URL provided by an Ingress object is changed by a user, would they then also have to change their RabbitmqCluster annotations to match?

Zerpet:
> One question I have; if the external URL provided by an Ingress object is changed by a user, would they then also have to change their RabbitmqCluster annotations to match?

That's correct. Automating this would imply that we start watching `Ingress` resources and do some correlation between `Ingress <-> RabbitmqCluster`. Such work could be possible, but perhaps as a second iteration?

docs/proposals/20220105-start-certificate-support.md
```diff
@@ -0,0 +1,269 @@
...
+
+![](../images/start-cert-internal-domain.jpg "Start Certificate Support with internal cluster domain")
+
+In certain scenarios, the above solution won't be sufficient. For example, RabbitMQ can be exposed via
+TCP Ingress to applications external to Kubernetes. In this scenario, RabbitMQ is provisioned with a
+certificate to match the FQDN used to connect to the Ingress, for example `rabbitmq-example.my-domain.com`.
+In this scenario, it is not possible to determine the URL from the pieces of information described above.
+In order to satisfy this scenario, the Topology Operator needs the URL as input, per RabbitMQ instance.
+The proposed solution is to use an annotation in `RabbitmqCluster`, with the URL to connect to as value.
+For example, the annotation `messaging.rabbitmq.com/override-url` with value `"https://myrabbit.example.com"`.
```
MirahImage, line 105: What about the solution where we continue to connect via `{protocol}://{service-name}.{namespace}.svc:{port}` but trust certificates with a different SAN? For example, we could choose to trust `myrabbit.example.com` despite connecting to `myrabbit.mynamespace.svc`.

MirahImage: I believe the cf routing release does this, since it's possible to register a route with a certificate with a different SAN, but I think that may also take advantage of controlling the entire routing system.

MirahImage (approved):

Issue #287, opened 2022-01-24
Topology Operator does not respect value of `management.path_prefix ` configuration


matbarker:
## Describe the bug

Declaring a user with the topology operator fails to declare with the error  "Error 404 (Object Not Found): Not Found"

## To Reproduce

```
apiVersion: v1
kind: Secret
metadata:
  name: user-secret
  namespace: xxx #same namespace as the deployed RabbitmqCluster
type: Opaque
stringData:
  username: test
  password: test
---
apiVersion: rabbitmq.com/v1beta1
kind: User
metadata:
  name: tutorial-user
  namespace: xxx #same namespace as the deployed RabbitmqCluster
spec:
  importCredentialsSecret:
    name: user-secret
  rabbitmqClusterReference:
    name: rabbitmq # name of the RabbitmqCluster
```
`kubectl apply -f .\user-repro.yaml`
```
secret/user-secret created
user.rabbitmq.com/tutorial-user created
```
` kubectl describe user.rabbitmq.com tutorial-user -n xxx`
```
Name:         tutorial-user
Namespace:    xxx
Labels:       <none>
Annotations:  <none>
API Version:  rabbitmq.com/v1beta1
Kind:         User
Metadata:
  Creation Timestamp:  2022-01-24T01:10:48Z
  Finalizers:
    deletion.finalizers.users.rabbitmq.com
  Generation:  2
  Managed Fields:
    API Version:  rabbitmq.com/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:importCredentialsSecret:
          .:
          f:name:
        f:rabbitmqClusterReference:
          .:
          f:name:
    Manager:      kubectl-client-side-apply
    Operation:    Update
    Time:         2022-01-24T01:10:48Z
    API Version:  rabbitmq.com/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .:
          v:"deletion.finalizers.users.rabbitmq.com":
      f:spec:
        f:rabbitmqClusterReference:
          f:namespace:
      f:status:
        .:
        f:conditions:
        f:credentials:
          .:
          f:name:
    Manager:         manager
    Operation:       Update
    Time:            2022-01-24T01:10:48Z
  Resource Version:  91619254
[...]

Issue #277, opened 2021-12-20
`User` or `Namespace` stuck on `Terminating` becuase `Secret` has been deleted already


carlosjgp:
## Describe the bug
`User` and `Permission` can't be deleted due to the finalizer when the `Namespace` is deleted without deleting all the containing resources first

## To Reproduce
1. Create NS and resources `kubectl apply -f`
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
---
apiVersion: v1
kind: Secret
metadata:
  name: test-creds
  namespace: test
type: Opaque
stringData:
  username: my-fav-user
  password: secure-pass
---
apiVersion: rabbitmq.com/v1beta1
kind: User
metadata:
  name: my-fav-user
  namespace: test
spec:
  tags:
  - management # available tags are 'management', 'policymaker', 'monitoring' and 'administrator'
  - policymaker
  rabbitmqClusterReference:
    name: rabbitmq
    namespace: test
  importCredentialsSecret:
    name: test-creds
---
apiVersion: rabbitmq.com/v1beta1
kind: Permission
metadata:
  name: my-fav-user
  namespace: test
spec:
  vhost: /
  user: my-fav-user
  permissions:
    write: ".*"
    configure: ".*"
    read: ".*"
  rabbitmqClusterReference:
    name: rabbitmq
    namespace: test
```

2. Delete the NS `kubectl delete ns test`
3. NS is stuck on `Terminating` state

2b. Delete `Secret` `kubectl delete secret -n test test-creds`
3b. Try to delete `User` `kubectl delete rabbitmq.com/v1beta1/users -n test test`
4b. `User` is stuck on `Terminating`

**Expected behavior**
NS is deleted and operator resources are managed correctly on the RabbitrMQ cluster, User and Permissions are deleted

## Version and environment information
- Messaging Topology Operator: `docker.io/bitnami/rmq-messaging-topology-operator:1.2.1-scratch-r0`
- RabbitMQ: `docker.io/bitnami/rabbitmq:3.8.26-debian-10-r25`
- RabbitMQ Cluster Operator: `docker.io/bitnami/rabbitmq-cluster-operator:1.10.0-scratch-r4`
- Kubernetes: `eks - 1.21`
- Cloud provider or hardware configuration: `aws`
 
## Additional context
[RabbitMQ operator Helm chart](https://artifacthub.io/packages/helm/bitnami/rabbitmq-cluster-operator) `2.0.8`

Logs:
[...]

PR #294, opened 2022-02-01
Add rabbitmqClusterReference.connectionSecret

### LINKED ISSUE #135: Be able to use existing RabbitMQ clusters that are not created/managed by https://github.com/rabbitmq/cluster-operator
Some users might want to decouple the creation of a rabbitmqcluster from using messaging-topology operator to manage resources on it. Think of it as "Bring my own RabbitMQ". Currently I do not see a way to do this, since all the existing resources require a reference to RabbitMQCluster, for example here:
https://github.com/rabbitmq/messaging-topology-operator/blob/main/api/v1beta1/binding_types.go#L43

And seems like we don't really need the rabbitmqcluster reference except to pull in the connection information so that we can create the client connection, so wondering if we could investigate providing a mechanism to be able to specify the connection information (say in a secret with predefined fields for example) so that we could support the use case for bringing your own existing RabbitMQ broker and not require that it be created via the RabbitMQCluster operator.
https://github.com/rabbitmq/messaging-topology-operator/blob/main/internal/rabbitmq_client_factory.go#L76

Perhaps instead of requiring that the credentials for the connection come from RabbitMQCluster reference:
https://github.com/rabbitmq/messaging-topology-operator/blob/main/api/v1beta1/binding_types.go#L43

We could add another way to specify required connection information to be able to connect to existing RabbitMQ clusters?

I don't currently see a way to support this use case because you must use a RabbitMQClusterReference.

**Additional context**

ChunyiLyu:
This closes #135 

## Summary Of Changes

Will be squashed merged due to messy history.
Needs to create a followup for handling TLS connection


api/v1beta1/exchange_webhook.go
```diff
@@ -19,8 +19,20 @@ func (r *Exchange) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+			field.Forbidden(field.NewPath("spec", "rabbitmqClusterReference"),
+				"do not provide both spec.rabbitmqClusterReference.name and spec.rabbitmqClusterReference.connectionSecret"))
+	}
+
+	if e.Spec.RabbitmqClusterReference.Name == "" && e.Spec.RabbitmqClusterReference.ConnectionSecret == nil {
+		return apierrors.NewForbidden(e.GroupResource(), e.Name,
+			field.Forbidden(field.NewPath("spec", "rabbitmqClusterReference"),
+				"must provide either spec.rabbitmqClusterReference.name or spec.rabbitmqClusterReference.connectionSecret"))
+	}
 	return nil
```
coro, line 36: Could we extract this test case into a common function used by all of the webhooks?

api/v1beta1/rabbitmq_cluster_reference.go
```diff
@@ -0,0 +1,42 @@
...
+	// Have to set either name or connectionSecret, but not both.
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// The namespace of the RabbitMQ cluster to reference.
+	// Defaults to the namespace of the requested resource if omitted.
+	// +kubebuilder:validation:Optional
+	Namespace string `json:"namespace,omitempty"`
+	// Secret contains the http management uri for the RabbitMQ cluster.
+	// The Secret must contain the key `uri` or operator will error.
+	// Have to set either name or connectionSecret, but not both.
```
coro, line 16: Could do with mentioning the username/password keys too

internal/rabbitmq_client_factory.go
```diff
@@ -54,21 +54,27 @@ var RabbitholeClientFactory RabbitMQClientFactory = func(rmq *rabbitmqv1beta1.Ra
 	return generateRabbitholeClient(rmq, svc, credsProvider, hostname, certPool)
 }
 
-// returns a http client for the given RabbitmqCluster
+// generateRabbitholeClient returns a http client for the given RabbitmqCluster
+// if provided RabbitmqCluster is nil, generateRabbitholeClient uses username, passwords, and uri
+// information from credsProvider to generate a rabbit client
 func generateRabbitholeClient(rmq *rabbitmqv1beta1.RabbitmqCluster, svc *corev1.Service, credsProvider CredentialsProvider, hostname string, certPool *x509.CertPool) (rabbitmqClient RabbitMQClient, err error) {
```
coro, line 60: I'm not sure I like the interface this function now provides. Seems like in order for code to call this function, it has to be tightly coupled to how it behaves under the covers. In this case, consuming code must know that rmq and svc can be nil, and that using the data that is provided in credsProvider depends on whether rmq is nil or not. It feels like the function is doing too much at once as a result.

I would try to make the factory functions as small as possible, and only take in what they need to construct the client. I would personally either move the logic for constructing the inputs to the factory either to the controllers, its own library, or separate functions in this library.

coro (commented):
Will test live now.

coro (commented):
One neat thing this provides is that it solves this issue: https://github.com/rabbitmq/messaging-topology-operator/issues/287

Zerpet:
> I'm seeing an issue where if I create a topology object, then the RabbitMQ cluster disappears for whatever reason, the object is stuck in failed delete forever as the operator hits tcp timeouts trying to connect.
> 
> In the case where we're using a LocalClusterReference, and we fail to find the RMQ, we assume it's been deleted and we delete the underlying topology object; I'd like to see a similar thing here potentially. There is a trade off, in that a cluster may only be temporarily unavailable while a user does a delete operation, and we'd leave some state around, but I think that's a better solution than having users do a `k delete --force`.

This behaviour is similar to what's described in #277. Big plus if this PR also fixes that problem.

docs/examples/rmqsourcecrd
```diff
@@ -0,0 +1,130 @@
+apiVersion: apiextensions.k8s.io/v1
```
coro, line 1: What is this file? Accidental checkin?

coro (approved):
Following on from our chat on Zoom, LGTM! Just one concern about an accidentally included file potentially?

docs/examples/rmqsourcecrd
ChunyiLyu: removed

PR #296, opened 2022-02-09
Support TLS for connectionSecret

### LINKED ISSUE #135: Be able to use existing RabbitMQ clusters that are not created/managed by https://github.com/rabbitmq/cluster-operator

ChunyiLyu:
This closes #135 

## Summary Of Changes

This adds TLS support for resources that are using `spec.rabbitmqClusterReference.connectionSecret`.
Certificates are provided in the same way as  `spec.rabbitmqClusterReference.name` use cases, user needs to either configure their RabbitMQ with public trusted certs, or mount CAs to the topology operator deployment.

To use secure TLS connection, users need to set scheme in the URI to `https` in provided connection secrets. When scheme is not provided, operator defaults to non-tls `http` scheme for connection.

## Follow up

Need to add docs to rabbitmq.com

### DIFF
--- a/internal/cluster_reference.go
+++ b/internal/cluster_reference.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
@@ -40,11 +41,7 @@ func ParseRabbitmqClusterReference(ctx context.Context, c client.Client, rmq top
 		if err := c.Get(ctx, types.NamespacedName{Namespace: requestNamespace, Name: rmq.ConnectionSecret.Name}, secret); err != nil {
 			return nil, false, err
 		}
-		creds, err := readCredentialsFromKubernetesSecret(secret)
-		if err != nil {
-			return nil, false, fmt.Errorf("unable to retrieve information from Kubernetes secret %s: %w", secret.Name, err)
-		}
-		return creds, false, nil
+		return readCredentialsFromKubernetesSecret(secret)
 	}
 
 	var namespace string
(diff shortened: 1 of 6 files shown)


Zerpet (commented):
Looking at the TLS system test in `system_tests/tls_system_test.go` makes we wonder if we should move this test at integration level, given that it doesn't need the running RabbitMQ to perform its assertions. This test is creating objects in Kubernetes, and asserting on other objects created/updated by the Top-Op, and this could be accomplished at integration level.

Do you know of any limitation/blocker that makes this test a must-be system level?

Everything else looks good 👍

internal/cluster_reference.go
```diff
@@ -133,18 +130,34 @@ func AllowedNamespace(rmq topology.RabbitmqClusterReference, requestNamespace st
...
+		return nil, false, fmt.Errorf("unable to retrieve information from Kubernetes secret %s: %w", secret.Name, errors.New("nil secret"))
+	}
+
+	uBytes, found := secret.Data["uri"]
+	if !found {
+		return nil, false, keyMissingErr("uri")
+	}
+
+	uri := string(uBytes)
+	if string(uBytes[0:4]) != "http" {
```
coro, line 144: Could we use `strings.HasPrefix`?

coro (approved):

ChunyiLyu:
@Zerpet IMO having the tls tests in the e2e system levels test the actual user workflow of mounting certs to topology operator, which we can't cover in an integration tests. I think this provides important coverage that otherwise would be missed in integration.

PR #310, opened 2022-02-18
Support for internal domain name as env var

Zerpet:
This is related #233 

## Summary Of Changes

- Support for custom internal cluster domain name
- Refactor unused code and extracted env var names as constants

## Additional Context

It's notable the testing strategy used to test this new feature. In order to test that a controller is using the property, I had to expose such property in the controller's struct, and add a function that allows to interact with such property. Secondly, the interface was necessary to avoid duplicating the same code multiple times, in every controller test. The addition of this interface seems to have no impact, and it allows us to test common controller functionality in one place.

There's been no work done to allow arbitrary URI strings, as that use case is better covered by #294 and #296.

Pending to PR the docs to document this new feature.


Makefile
```diff
@@ -8,6 +8,10 @@ platform := $(shell uname | tr A-Z a-z)
 list:    ## list Makefile targets
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 
+### Tools
+# Allows flexbility to use other build kits, like nerdctl
+BUILD_KIT ?= /usr/local/bin/docker
```
coro, line 13: Love it!

main.go
```diff
@@ -42,6 +45,25 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+func sanitizeClusterDomainInput(clusterDomain string) string {
+	if len(clusterDomain) == 0 {
+		return ""
+	}
+
+	match, _ := regexp.MatchString("^\\.?[a-zA-Z-]+(\\.[a-zA-Z-]+)*$", clusterDomain) // Allow-list expression
```
MirahImage, line 53: This regex is not general enough, for example, it doesn't allow numbers within domain names.

MirahImage (changes requested):

main.go
Zerpet: Good catch! I'll build an expression to match following the [RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.1), except the 63 character limitation per label (hostname, subdomain, etc.) since it'll make a bit awkard to read to the regex, and it should be validated by Kubernetes itself, during the cluster initialisation process.

controllers/common_test.go
```diff
@@ -0,0 +1,169 @@
...
+		fakeRabbitMQClient.DeleteShovelReturns(commonHttpDeletedResponse, nil)
+	})
+
+	It("sets the domain name in the URI to connect to RabbitMQ", func() {
+		for _, controller := range topologyControllers {
+			controller.SetInternalDomainName(".some-domain.com")
+		}
+
+		for i, _ := range topologyControllers {
+			Expect(client.Create(ctx, topologyObjects[i])).To(Succeed())
```
Zerpet, line 117: This test was a bit flaky. For some reason, sometimes the test would fail by not adding the internal domain name. By separating in two different loop the "setup" and the "tests", the flake seems to go away.

MirahImage (approved):

coro (approved):

Issue #324, opened 2022-03-01
User and Permission stuck in Terminating


LucasBoisserie:
## Describe the bug
We use ArgoCD to deploy our code in Kubernetes. We create/delete multiple review-apps each hour, they include: exchanges, queues, permissions and users.
We notice that sometimes User and Permission are stuck in Terminating phase.

In the operator log, we found the following error message: `failed to retrieve user credentials secret from status`

## To Reproduce
1. Create User
2. Create Permission with userReference
3. Remove User
4. Remove Permission (stuck in Terminating phase)

We notice sometimes than step 3 stay stuck also in Terminating phase.

```
---
# User.yaml
apiVersion: rabbitmq.com/v1beta1
kind: User
metadata:
  finalizers:
  - deletion.finalizers.users.rabbitmq.com
  labels:
    argocd.argoproj.io/instance: sga-qualif-rec
  name: user-sga-hub-rec
  rabbitmqClusterReference:
    name: rabbitmq-qualif
    namespace: rabbitmq-qualif
---
#Permission.yaml
apiVersion: rabbitmq.com/v1beta1
kind: Permission
metadata:
  finalizers:
  - deletion.finalizers.permissions.rabbitmq.com
  labels:
    argocd.argoproj.io/instance: sga-qualif-rec
  name: permission-hub-sga-rec
spec:
  permissions:
    configure: .*
    read: .*
    write: .*
  rabbitmqClusterReference:
    name: rabbitmq-qualif
    namespace: rabbitmq-qualif
  userReference:
    name: user-sga-hub-rec
  vhost: vhost-hub-rec
```

**Expected behavior**
All components are removed

## Version and environment information
- Messaging Topology Operator: [1.2.1-scratch-r0] (deployed with bitnami chart)
- RabbitMQ: [3.8.23-debian-10-r35]
- RabbitMQ Cluster Operator: [1.11.0-scratch-r0] (deployed with bitnami chart)
- Kubernetes: [e.g. 1.22.7]
- Cloud provider : Scaleway Kapsule
 
## Additional context
After some investigations into the logs and the source code of the operator, we find two issues:
  
### Permission with userReference stuck in Terminating if User is removed first 

If the User is removed before the Permission, the Permission can’t be removed because the operator can’t found rabbitmq username stored inside User secret credentials.(https://github.com/rabbitmq/messaging-topology-operator/blob/main/controllers/permission_controller.go#L63)
It throws an error and retriggers the object.

[...]

PR #325, opened 2022-03-01
Update logic deletion on user and permission controller

LucasBoisserie:
Ref #324 

## Summary Of Changes

In permission controller:
- add owner reference from User on Permission when userReference is set
- add condition: when the User doesn’t exist and Permission is in deletion status, we skip the permission removal

In user controller:
- set in the User status part a field with the real username to not depend on the secret

### DIFF
--- a/api/v1beta1/user_types.go
+++ b/api/v1beta1/user_types.go
@@ -44,6 +44,8 @@ type UserStatus struct {
 	Conditions         []Condition `json:"conditions,omitempty"`
 	// Provides a reference to a Secret object containing the user credentials.
 	Credentials *corev1.LocalObjectReference `json:"credentials,omitempty"`
+	// Provide rabbitmq Username
+	Username string `json:"username"`
 }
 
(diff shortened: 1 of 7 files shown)


Zerpet (commented):
Hey there, I've finally found sometime to go over this, thank you for your patience!

The code changes look good to me and some quick testing verified the compatibility with previous versions of the CRD. The only thing I'm missing is a test case in the controllers that catches this bug, so that we catch in the future in the undesirable case of a regression. Although I can't quite suggest a good test case, given that in the controller test suite, we can't know if a resource is not deleted because `testenv` slowness, or because the reconcile is failing 🤔 

Let me consult the team regarding the test case. Feel free to suggest a test case if you have any ideas 🙂

LucasBoisserie:
@Zerpet, I added some tests on Permissions and Users controller like you suggest
I also split the `if `

Zerpet (approved):
Latest changes look good to me. Thank you!

PR #342, opened 2022-03-24
Validate Quorum queue settings in WebHook

Zerpet:

## Summary Of Changes
We were allowing non-durable quorum queues as a valid manifest. This results in a 400
response from RabbitMQ server. The error is not visible to the user,
until they describe the queue and observe the status condition. Now it
fails sooner, which is good given that we know these settings will never
succeed.


ChunyiLyu (approved):
Great!

PR #350, opened 2022-04-14
Temporary fix on deletion_system_test in order to avoid to create a m…

DanielePalaia:
…irror queue in a one pod cluster causing the pod to get stuck in "terminating" phase and preventing our pipeline to work.

It can be reverted once await_online_synchronized_mirror get fixed to manage this case

Not really a temporary fix anymore, putting grace period to a lower value will increase the stability of this test in general, as the scope is to delete objects when the rabbimq cluster get deleted.

This closes #

### DIFF
--- a/system_tests/deletion_system_test.go
+++ b/system_tests/deletion_system_test.go
@@ -45,7 +45,7 @@ var _ = Describe("Deletion", func() {
 			},
 			Spec: topology.PolicySpec{
 				Name:    "policy-deletion-test",
-				Pattern: ".*",
+				Pattern: "not-match-queue",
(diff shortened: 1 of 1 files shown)


coro (approved):

Zerpet (changes requested):
I'd much rather mark the test as "skip" than giving us a false sense of successful test.

I think the test has caught a legitimate bug, in the Core in this case. If we want to make the pipeline green, we should just skip this test.

DanielePalaia:
No, that's of course fine! Done!

Zerpet (approved):
Ship it!

PR #358, opened 2022-05-02
Schema replication example

DanielePalaia:
This closes #
Added an example to use the schema replication plugin using the topology operator part of:
https://github.com/rabbitmq/service-operator-experience/issues/46

### DIFF
--- a/docs/examples/schema-replication/README.md
+++ b/docs/examples/schema-replication/README.md
@@ -6,8 +6,12 @@ This feature is not available in the open source RabbitMQ distribution.
 
 Schema Replication can be configured using the topology operator.
 
-In the upstream-rabbitmq.yaml and downstream-rabbitmq.yaml files are present the configuration definitions in order to setup two RabbitMQ clusters with the schema replication plugin enabled and relative prerequisites (user, vhost, permissions).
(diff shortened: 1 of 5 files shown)


Zerpet (commented):
It looks well overall. Here are a few points that we may revisit to improve the experience of using this example:

- Using a load balancer service type is not always possible e.g. kind
- This example requires manual editing the `schema-replication-*` files with a real IP
- Namespaces `upstream` and `downstream` must already exist

If we deploy both `RabbitmqCluster` in the same cluster, we can leverage Kubernetes internal DNS. Then we can declare the endpoints as `<instance-name>.<namespace>.svc` i.e. `upstream.upstream.svc:5672`. This addresses the first two pain points.

Regarding the namespaces, we could include a `Namespace` object in the YAML manifests, or make a comment in the readme.

For the purpose of demonstrating the capabilities of the plugin and the Topology Operator, deploying in the same Kubernetes cluster and different namespaces is acceptable.

DanielePalaia:
@Zerpet this is fine. I modified the endpoints of the SchemaReplication and I also renamed upstream and downstream clusters in upstream-cluster and downstream-cluster in order to differentiate the names from the namespaces.
For the first point, as this should run on a commercial versions the users needs also to setup secrets for the image registry after creating the namespaces, so I documented it in the README. What do you think?

DanielePalaia:
Closed as it was discussed to just document it in the commercial version.

PR #369, opened 2022-05-12
Support setting manager sync period using operator envVar

### LINKED ISSUE #243: Resource deleted in admin interface is not recreated before restarting operator
## Describe the bug
Manually deleted resources are not reconciled before operator restart

## To Reproduce
1. deploy a resource ex a policy or a user
2. login with default user and delete resource
3. The resource do not come back

4. restart topology-operator pod. 
5. resources are reconciliation again

**Expected behavior**
if a resource is deleted manually from management console, next reconciliation loop from operator should re-add the resources.
If there is no loop, i guess this is a feature request for adding a timer to run reconciliation once per x min. maybe 5min. 

## Version and environment information
- Messaging Topology Operator: 1.1.0
- RabbitMQ: [e.g. 3.8.22]
- RabbitMQ Cluster Operator: 1.8.3
- Kubernetes: 1.20.7

ChunyiLyu:
This closes #243

## Summary Of Changes

- manager sync period configuration is for all controllers under the same manager
- SYNC_PERIOD won't have a default, user need to op in for this behavior as we cannot determine a reasonable default
- SYNC_PERIOD needs to be in a format that's parsable by time.ParseDuration(), else exit 1. When successfully parsed, the operator logs at starting, e.g.`{"level":"info","ts":1652354704.2477453,"logger":"setup","msg":"sync period set; all resources will be reconciled every: 5m0s"}`
- can be updated like any other env var

## Additional Context

[same functionality in crossplane](https://github.com/crossplane/crossplane/blob/1344a86018f03fcbba6b9365c63696dcd47ebcf6/cmd/crossplane/core/core.go#L71)
[discussion in controller runtime](https://github.com/kubernetes-sigs/controller-runtime/issues/521)

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -39,6 +39,7 @@ const (
 	KubernetesInternalDomainEnvVar = "MESSAGING_DOMAIN_NAME"
 	OperatorNamespaceEnvVar        = "OPERATOR_NAMESPACE"
 	EnableWebhooksEnvVar           = "ENABLE_WEBHOOKS"
+	ControllerSyncPeriodEnvVar     = "SYNC_PERIOD"
 )
 
 type TopologyController interface {
--- a/main.go
+++ b/main.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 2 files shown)


Zerpet (approved):
Beautiful!! Don't forget to PR the docs, or open an issue to PR the docs, with this new feature 📖

PR #381, opened 2022-05-27
Extract generating rmq client into its own package

ChunyiLyu:
This closes #

## Summary Of Changes

- so we can start reusing in other operators
- some renaming of functions so the package name and func names won't repeat themselves


Zerpet (approved):
Very nice! 🚀

PR #384, opened 2022-06-01
Update vulnerable dependency

Zerpet:

## Summary Of Changes

GoYaml dependency has a known CVE-2022-28948. It is fixed in versions
v3.0.0+

Go Client also has a dependency on this package. It looks like Client Go
does not use the vulnerable version.

Regenerated Client Set since we changed our dependencies.

## Additional Context

Unit and integration tests are passing locally.


ChunyiLyu (approved):
YES

coro (approved):

PR #387, opened 2022-06-06
Support referencing a vault secret path to configure schema sync credentials

### LINKED ISSUE #386: schemareplication cr should support getting username/password from Vault

 As part of working on https://github.com/rabbitmq/service-operator-experience/issues/64, I found users still need to provide a k8s secret that has username/password for a rmq to be able to set schema sync parameters. This behavior should be updated such that the schema sync CR can take a vault secret path to find username and password for a rabbitmqcluster that uses vault as secret backend. The same should be true for offsitereplication CR to fully support vault.

Users can provide a path to rabbitmq credentials in vault instead of a k8s secret to configure schema sync CR.
Should be an addition to the CRD and not impact existing users.

ChunyiLyu:
This closes #386

## Summary Of Changes

Users will be able to reference a vault secret path `spec.secretBackend.vault.secretPath` to configure username and password for schema sync parameters.

- require either spec.secretBackend.vault.secretPath or spec.secretBackend.vault, but not both (verified in webhook)
- tested in integration tests suite (and manually verified behavior on create/update/delete)
- `spec.secretBackend.vault.secretPath` is updatable
- example schemasync.rabbitmq.com using vault
```yaml
---
kind: SchemaReplication
metadata:
  name: upstream
spec:
  endpoints: "test:134556"
  secretBackend:
    vault:
      secretPath: secret/data/rabbitmq/config
  rabbitmqClusterReference:
    name: vault-default-user
```


api/v1beta1/schemareplication_webhook.go
```diff
@@ -36,10 +41,25 @@ func (s *SchemaReplication) ValidateUpdate(old runtime.Object) error {
...
 }
+
+func (s *SchemaReplication) validateSecret() error {
+	if s.Spec.UpstreamSecret != nil && s.Spec.UpstreamSecret.Name != "" && s.Spec.SecretBackend.Vault != nil && s.Spec.SecretBackend.Vault.SecretPath != "" {
+		return apierrors.NewForbidden(s.GroupResource(), s.Name,
+			field.Forbidden(field.NewPath("spec"),
+				"do not provide both secretBackend.vault.secretPath and upstreamSecret"))
+	}
+
+	if s.Spec.UpstreamSecret == nil && s.Spec.SecretBackend.Vault == nil {
```
coro, line 59: This doesn't catch Vault being non-nil but with the secret unset, e.g.:

```yaml
---
kind: SchemaReplication
metadata:
  name: upstream
spec:
  endpoints: "test:134556"
  secretBackend:
    vault: {}
  rabbitmqClusterReference:
    name: hello-world
```

This is later caught in the controller, but it's not validated upfront by this webhook.

coro (approved):
LGTM! I've one case where some invalid YAML can make it through to the controller; I'll leave it up to you if you would like to fix it or not.

api/v1beta1/schemareplication_webhook.go
ChunyiLyu: Good point and updated.

PR #389, opened 2022-06-07
Rename ValidatingWebhookConfiguration to avoid conflicts from other c…

evankanderson:

## Summary Of Changes

- Change kustomization for webhook so that the generated webhook doesn't collide with webhooks for other projects. This was noticed as a conflict with `cartographer`, but could happen with _any_ other Kubernetes controller based on controller-runtime.
- Note that this doesn't change the Service name for the webhook, because the Service is namespaced, and so won't collide with other components installed in the cluster. The ValidatingWebhookConfiguration is cluster-scoped, so the namespace collision is more of a concern (particularly since `controller-gen` generates that name by default).

## Additional Context

- You probably need some install instructions about how to clean up the existing webhook; this seems worth it given the likelihood of future collisions.

### DIFF
--- a/config/webhook/kustomization.yaml
+++ b/config/webhook/kustomization.yaml
@@ -4,3 +4,14 @@ resources:
 
 configurations:
 - kustomizeconfig.yaml
+
+patches:
+- patch: |-
+    - op: replace
+      path: /metadata/name
+      value: topology.rabbitmq.com
+  target:
+    group: admissionregistration.k8s.io
+    version: v1
+    kind: ValidatingWebhookConfiguration
+    name: .*


ChunyiLyu:
Correction from the previous comment https://github.com/rabbitmq/messaging-topology-operator/pull/388#issuecomment-1148399908

Patch is correctly applied with `make generate-manifests` and other deploy make targets. As for the svc, the team could  update webhook service name in a different PR.

Zerpet (changes requested):
Hey there 👋 Thank you for the contribution! I'm not sure what's up with the CLA bot, but given your commit is signed with a VMware email, I'm happy to just merge this this PR when it's ready. 
On that front, I have a change to request regarding the commit message. The subject line of your commit message is longer than 50 characters, which does not adhere to our [contribution guidelines](https://github.com/rabbitmq/messaging-topology-operator/blob/main/CONTRIBUTING.md#pull-requests), specifically: _commit with a [descriptive message](https://cbea.ms/git-commit/#seven-rules)_.

I'll be happy to merge this once you amend the commit to adhere to the guidelines. Thank you!

evankanderson:
Sorry for the delay, I was out for a bit.

I removed the justification part of the change from the first line, leaving only the description of what changed.

Zerpet (approved):
Thank you!

PR #390, opened 2022-06-08
Openshift integration

DanielePalaia:
Adding finalizers permissions in clusterrole definition

This closes #

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -45,6 +45,12 @@ rules:
 - apiGroups:
   - rabbitmq.com
   resources:
+  - bindings/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - rabbitmq.com
+  resources:
   - bindings/status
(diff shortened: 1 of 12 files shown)


ChunyiLyu (changes requested):
I think `update` as one verb should be enough for finalizer management. This is what we have for cluster operator: https://github.com/rabbitmq/cluster-operator/blob/main/controllers/rabbitmqcluster_controller.go#L82

Could you please take a look what's the minimal set of verb we need instead of adding all of them? Thanks!

DanielePalaia:
@ChunyiLyu I made some tests and yes update should be enough! Fixed the commit thanks!

Zerpet (approved):
Looks good with the new changes. I'll wait for @ChunyiLyu to confirm and give approval ✅

ChunyiLyu (approved):
Thank you for updating 😺

PR #396, opened 2022-06-15
Add debug pprof endpoints to topology operator

coro:

## Summary Of Changes
See https://github.com/rabbitmq/cluster-operator/pull/1064


DanielePalaia (approved):

Zerpet (commented):
Given the dependency bump on cluster operator, I believe you need to update the following file to make the integration tests happy again:

https://github.com/rabbitmq/messaging-topology-operator/blob/2a1b7b3e9862c44261d23603e214f5c44310de1e/controllers/suite_test.go#L80-L85

PR #403, opened 2022-06-20
openshift integration

DanielePalaia:
Wrap expect into eventually on updating status condition in system tests
to let tests be more stable in the smaller openshift environment

This closes #

### DIFF
--- a/system_tests/tls_system_test.go
+++ b/system_tests/tls_system_test.go
@@ -141,7 +141,7 @@ var _ = Describe("RabbitMQ Cluster with TLS enabled", func() {
 		Eventually(func() []topology.Condition {
 			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: policy.Name, Namespace: policy.Namespace}, &fetchedPolicy)).To(Succeed())
 			return fetchedPolicy.Status.Conditions
-		}, 10, 2).Should(HaveLen(1))
+		}, waitUpdatedStatusCondition, 2).Should(HaveLen(1), "policy status condition should be present")
 
 		readyCondition := fetchedPolicy.Status.Conditions[0]
 		Expect(string(readyCondition.Type)).To(Equal("Ready"))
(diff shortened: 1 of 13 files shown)


coro (changes requested):
Is the intermittent failure due to the Status Condition not being present, or having the wrong value? As of this change, the Eventually will loop until a status condition is returned, regardless of if it's 'True' or not, and then still fail if it's 'False'. Perhaps the checks on each returned condition should be part of the Eventually loop?

Zerpet (commented):
I would add a descriptive message for the `Should` assertion, so that Gomega reports something more useful than "Expected 1, received 0" 🙂 
https://onsi.github.io/gomega/#annotating-assertions

DanielePalaia:
@coro Actually in this case the condition seems not to be present, as the error result is:  <[]v1beta1.Condition | len:0, cap:0>: nil, but after putting a timeout seems then to always be present.

DanielePalaia:
@ChunyiLyu @Zerpet Thanks for the suggestions. Implemented!

system_tests/tls_system_test.go
```diff
@@ -141,7 +141,7 @@ var _ = Describe("RabbitMQ Cluster with TLS enabled", func() {
 		Eventually(func() []topology.Condition {
 			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: policy.Name, Namespace: policy.Namespace}, &fetchedPolicy)).To(Succeed())
 			return fetchedPolicy.Status.Conditions
-		}, 10, 2).Should(HaveLen(1))
+		}, 20, 2).Should(HaveLen(1), "policy status condition should be present")
```
ChunyiLyu, line 144: I think you would like to use `waitUpdatedStatusCondition`here as well

DanielePalaia: Yes indeed :) fixed, thanks!

system_tests/utils.go
```diff
@@ -27,6 +28,9 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+// Used for openshift smaller cluster
```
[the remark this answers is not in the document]
DanielePalaia: Fair enough!

coro (approved):

PR #423, opened 2022-07-26
Add support to read path prefix from cluster specs

### LINKED ISSUE #421: management.path_prefix is not respected by messaging-topology-operator
## Describe the bug
Declaring a user with the topology operator fails to declare with the error "Error 404 (Object Not Found): Not Found" when the management.path_prefix is set to some value (that is needed to expose the UI with reverse proxy, ingress etc) in the RabbitmqCluster spec at .spec.rabbitmq.additionalConfig

As a workaround using [connection secret ](https://www.rabbitmq.com/kubernetes/operator/using-topology-operator.html#non-operator) was offered in #287  but this needs a user/password credential that is already present in the cluster thus creating a "chicken or egg" problem when one wants to set up a cluster in fully declarative way using gitops (argocd)

## To Reproduce
1, set management.path_prefix RabbitmqCluster spec at .spec.rabbitmq.additionalConfig to some value
2, create user as shown in [example](https://github.com/rabbitmq/messaging-topology-operator/blob/main/docs/examples/users/publish-consume-user.yaml)
3, check logs of messaging-topology-operator pod in rabbitmq-system

**Expected behavior**
As @blackjid suggested https://github.com/rabbitmq/messaging-topology-operator/issues/287#issuecomment-1112502040 The topology operator should look into the cluster reference resource to see if there is a path_prefix config, and use it to define the api calls. Maybe the Cluster resource could have a specific key in the spec to add a management path prefix, so it would be easier for the topology operator to find the information.

## Version and environment information
- Messaging Topology Operator:  1.6.0
- RabbitMQ: 3.9.13
- RabbitMQ Cluster Operator: 1.12.1
- Kubernetes: 1.19.10
- Cloud provider or hardware configuration: onprem

blackjid:
This closes #421

## Summary Of Changes

This pr add support for the operator to look at the cluster specs for the `path_prefix` attribute if it is set. This way the operator can communicate with the cluster api while it can be exposed on a different path.

## Additional Context

The `additional_config` property on the cluster spec adds configuration to for the cluster.
I understand the cluster support ini and sysctl like configurations. 

- [x] read configuration from the clusterspec
- [x] `make unit-tests` PASS

ref: 
https://www.rabbitmq.com/kubernetes/operator/using-operator.html#additional-config
https://www.rabbitmq.com/configure.html#config-file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/onsi/gomega v1.20.0
 	github.com/rabbitmq/cluster-operator v1.14.0
 	github.com/sclevine/yj v0.0.0-20200815061347-554173e71934
+	gopkg.in/ini.v1 v1.66.6
 	k8s.io/api v0.24.3
 	k8s.io/apimachinery v0.24.3
 	k8s.io/client-go v0.24.3
--- a/go.sum
+++ b/go.sum
@@ -1966,6 +1966,7 @@ gopkg.in/inf.v0 v0.9.1 h1:73M5CoZyi3ZLMOyDlQh031Cx6N9NDJ2Vvfl76EDAgDc=
 gopkg.in/inf.v0 v0.9.1/go.mod h1:cWUDdTG/fYaXco+Dcufb5Vnc6Gp2YChqWtbxRZE0mXw=
 gopkg.in/ini.v1 v1.51.0/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
 gopkg.in/ini.v1 v1.62.0/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
+gopkg.in/ini.v1 v1.66.6 h1:LATuAqN/shcYAOkv3wl2L4rkaKqkcgTBQjOyYDvcPKI=
 gopkg.in/ini.v1 v1.66.6/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
 gopkg.in/natefinch/lumberjack.v2 v2.0.0 h1:1Lc07Kr7qY4U2YPouBjpCLxpiyxIVoxqXgkXLknAOE8=
 gopkg.in/natefinch/lumberjack.v2 v2.0.0/go.mod h1:l0ndWWf7gzL7RNwBG7wST/UCcT4T24xpD6X8LsfU/+k=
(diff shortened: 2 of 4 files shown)


blackjid:
Hi @coro, well, I think it should be working.
There are still a few things to work on. But I think it might be a good idea to just expose it for feedback :)

blackjid:
I'd love to have some feedback on the way to support both `ini` and `sysctl` config from the cluster.

### For the `ini` syntax

Using `go-ini/ini` package works very nice to read the `ini` syntax. I think this is ok.

### For the `sysctl` like syntax

I have been trying this, I have to approaches.

1. I haven't found an open library to parse this files, but the specification is pretty simple so I've implemented simple parser that is working fine in the tested scenarios.
2. I just find out that I could us the `go-ini/library` to parse the `sysctl` config, as it supports empty sections.

    Instead of doing this to read the value from a ini config

    ```
    cfg.Section("management").Key("path_prefix).String()
    ``` 
    I could do this from a sysctl config
    ```
    cfg.Section("").Key("management.path_prefix").String()
    ```

    The specification from the ini config is very similar to the sysctl, minus the sections.


What do you thing would be better?

coro:
Ah, there may be confusion from our documentation. RabbitMQ doesn't support the true .ini config format, only the sysctl format from a library called [cuttlefish](https://github.com/basho/cuttlefish). The only rules of the syntax are:

* Everything you need to know about a single setting is on one line
* Lines are structured Key = Value
* Any line starting with # is a comment, and will be ignored.

(From https://github.com/basho/cuttlefish/wiki/Cuttlefish-for-Application-Users)

So in summary, you don't need to worry about .ini, only sysctl. The go-ini library you've chosen is great, and is what we use in the cluster-operator to [set and parse the additional config files](https://github.com/rabbitmq/cluster-operator/blob/6d5295b4858c3b70b6ad3cf63dee766f9e3a00b1/internal/resource/configmap.go#L98-L213) originally. Of your options for sysctl parsing above, I prefer the last one because it matches how the keys are set originally:

```
cfg.Section("").Key("management.path_prefix").String()
```

blackjid:
Thanks for the feedback... that was exactly what I needed :)
I updated the PR with the latest changes, let me know if there is something else to add or change.

coro (approved):
LGTM! Let me pass this to another member of our team for a second opinion.

ChunyiLyu (approved):
Thank you for contributing!

blackjid:
> Thank you for contributing!

No problem!, I'm really happy to do it :)

Is there any plan of release this?

ChunyiLyu:
@blackjid New release just cut 😊 See: https://github.com/rabbitmq/messaging-topology-operator/releases/tag/v1.8.0

blackjid:
Awesome! Thanks

PR #432, opened 2022-08-19
Small tests updates

ChunyiLyu:

## Summary Of Changes

Some small tests updates and regenerate CRDs after k8s api and controller-gen was bumped.

komega is a test client from [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/komega/komega.go). Can be used in more places in integration tests but just the finalizer tests for now.


MirahImage (approved):
Huge diff, very minor changes 😂 
Looks good.

PR #456, opened 2022-09-21
Topic permission 191

### LINKED ISSUE #191: Be able to create/update/delete topic permissions
```gherkin
As a service operator running RabbitMQ on Kubernetes
I would like to declaratively configure topic permissions
So that I can enforce authorization for MQTT and STOMP applications
```

An API for topic permissions, similar to the API for [permissions](https://github.com/rabbitmq/messaging-topology-operator/blob/main/config/crd/bases/rabbitmq.com_permissions.yaml) or [policies]((https://github.com/rabbitmq/messaging-topology-operator/blob/main/config/crd/bases/rabbitmq.com_policies.yaml))

As with other similar APIs, the contract with the cluster-operator should be minimal, via the cluster-reference and consuming the default-user secret, which conforms to the [Provisioned Service binding specification](https://k8s-service-bindings.github.io/spec/#provisioned-service).
The controller reconciling the Custom Resources should interact with the RabbitMQ cluster via the [RabbitMQ HTTP API](https://www.rabbitmq.com/management.html#http-api) via the kubernetes service associated to the RabbitMQ cluster.

**Additional context**
Topic authorisation targets protocols like STOMP and MQTT, which are structured around topics and use topic exchanges under the hood. RabbitMQ supports topic authorisation for topic exchanges. The routing key of a message published to a topic exchange is taken into account when publishing authorisation is enforced (e.g. in RabbitMQ default authorisation backend, the routing key is matched against a regular expression to decide whether the message can be routed downstream or not). 

Topic authorisation is an additional layer on top of existing checks for publishers. Publishing a message to a topic-typed exchange will go through both the basic.publish and the routing key checks. The latter is never called if the former refuses access.

Topic authorisation can also be enforced for topic consumers. Note that it works differently for different protocols. The concept of topic authorisation only really makes sense for the topic-oriented protocols such as MQTT and STOMP. In AMQP 0-9-1, for example, consumers consume from queues and thus the standard resource permissions apply. In addition for AMQP 0-9-1, binding routing keys between an AMQP 0-9-1 topic exchange and a queue/exchange are checked against the topic permissions configured, if any.

ChunyiLyu:
This closes #191

## Summary Of Changes

- add CRD `topicpermissions.rabbitmq.com` and tested in integration and system tests
- supports create/update/delete
- `vhost`, `exchange`, and `user` information are immutable and required
- since the logic to retrieve username from `spec.userReference` is the same as what's in `PermissionReconciler`, method  `r.getUserFromReference` was refactored to `getUsernameFromUser` to reuse in topic permission reconciler.

Will squash when merging to clean up history.


api/v1beta1/topicpermission_types.go
```diff
@@ -0,0 +1,83 @@
+package v1beta1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// TopicPermissionSpec defines the desired state of TopicPermission
+type TopicPermissionSpec struct {
```
DanielePalaia, line 10: We noticed that TopicPermissionSpec struct is identical to PermissionSpec of the Permission type. It would be possible to reuse it or maybe is good to have them separate for future extensions?

api/v1beta1/topicpermission_webhook_test.go
```diff
@@ -0,0 +1,144 @@
...
+			Expect(apierrors.IsInvalid(invalidPermission.ValidateCreate())).To(BeTrue())
+		})
+		It("does not allow both user and userReference to be unset", func() {
+			invalidPermission := permission.DeepCopy()
+			invalidPermission.Spec.UserReference = nil
+			invalidPermission.Spec.User = ""
+			Expect(apierrors.IsInvalid(invalidPermission.ValidateCreate())).To(BeTrue())
+		})
+
+		It("does not allow both spec.rabbitmqClusterReference.name and spec.rabbitmqClusterReference.connectionSecret be configured", func() {
```
DanielePalaia, line 44: We were noticing that there are not reference of "connectionSecret" in topic_permission_webhook.go, where is this situation managed?

config/samples/rabbitmq.com_v1beta1_topicpermission.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: rabbitmq.com/v1beta1
```
DanielePalaia, line 1: Should this file be here? We don't see the folder created on the repo.

DanielePalaia (approved):
This looks good! Thank you very much. it looks in line with the other controllers. We just noticed that there is a file in samples/rabbitmq.com_v1beta1_topicpermission.yaml that shouldn't be there. From Daniele and @ablease!

api/v1beta1/topicpermission_webhook_test.go
ChunyiLyu: It's validated as a method on `RabbitmqClusterReference` so it's shared across all webhooks: https://github.com/rabbitmq/messaging-topology-operator/blob/topic-permission-191/api/v1beta1/topicpermission_webhook.go#L41

api/v1beta1/topicpermission_types.go
ChunyiLyu: Structure of their the permission configurations are different: https://github.com/rabbitmq/messaging-topology-operator/blob/topic-permission-191/api/v1beta1/permission_types.go#L30-L37

User permissions needs `configrue` `read` and `write`, whereas topic permission is name of the exchange, `read` and `write`.

PR #522, opened 2023-01-06
add version option to expose the application version information

cpanato:

## Summary Of Changes

- add a version option to expose the application version information

This is useful when you need to verify which version the application is running in a quick way, it also logs and can be used to export a metric to prometheus (when have)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,8 +16,9 @@ COPY controllers/ controllers/
 COPY internal/ internal/
 COPY rabbitmqclient/ rabbitmqclient/
 
+ARG LDFLAGS=""
 # Build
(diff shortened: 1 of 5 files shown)


DanielePalaia:
Hi @cpanato thank you for your PR. It is easily possible to get the operator version and from it get the tag and the commit directly from github. It seems like this can be useful for local testing or debugging. Can you give us more information/use case where it can be useful?

cpanato:
@DanielePalaia was more in the case that we deploy using the digest and not the tag itself and to be easier to fetch the version without need to go back to github.
but i agree with you the benefit is small for a broader use cases
will close this for now

PR #532, opened 2023-01-18
Clean up integration tests

ChunyiLyu:
This closes #

## Summary Of Changes

This PR deletes 35 test specs from integration tests (125 -> 70 specs). 

It’s been painful to iterate integration tests. The integration tests has 125 specs, which takes about 6 to 10 minutes to finish (slower than system tests). It’s flaky because most tests wait for event and status conditions to be updated which takes a long time when lots of CRDs are created against a env test server.

For most of controllers, their integration tests looks roughly (with exception for one or two controllers that verify other things):

```
* a successful creation, where we assert on finalizer and status condition
* a failed create due to a http error, where we assert on status condition
* a failed create due to a go error, where we assert on status condition

* a successful delete, where we assert on published event
* a failed delete due to a http error, where we assert on published event
* a failed delete due to a go error, where we assert on published event

* a failed create due to referencing a prohibited namespace, where we assert on status condition
* a successful create for creating in an allow-all namespace
* a successful create for creating in an approve-listed namespace
```

This PR:
- removes successful create and delete cases that verify status conditions because they are covered in system tests
- removes all namespace permission test specs. Function 'ParseReference' from package `rabbitmqclient` performs namespaces permission checks. These tests are more suited to in `rabbitmqclient`
- asserts correct finalizer in system tests for CRs since they are removed from integration tests as part of removing the successful create test specs
- adds the same set of tests about namespace permission to `rabbitmqclient package`

## Does that mean integration tests are not going to flake anymore merging this PR?

Sadly no. With 70 test specs and still lots of eventually waits on status condition and published events, test flakes will happen. However, it will be less likely to and operator will have less duplicated test codes.


rabbitmqclient/cluster_reference_test.go
```diff
@@ -534,6 +533,142 @@ var _ = Describe("ParseReference", func() {
...
+		})
+
+		When("there is a list of allowed namespaces", func() {
+			BeforeEach(func() {
+				existingRabbitMQCluster.ObjectMeta.Annotations = map[string]string{
+					"rabbitmq.com/topology-allowed-namespaces": "allowed1,allowed2",
+				}
+				objs = []runtime.Object{existingRabbitMQCluster, existingCredentialSecret, existingService}
+			})
+			When("requested namespace is in topology-allowed-namespaces annotation", func() {
```
Zerpet, line 607: I'd prefer not having implementation details in the test description. As of now, this test reads as "context namespace permissions, when there is a list of allowed namespaces, and when requested namespace is in topology-allowed-namespaces annotation, it works". There is an implementation detail (annotation name) in the description. This implementation detail is not necessary to explain the behaviour.

IMO, it would read better as "context namespace permissions, when there is a list of allowed namespaces, and when requested namespace is allowed, it works". In this way, we would describe the scenario without tying its description to the annotation name.

rabbitmqclient/cluster_reference_test.go
```diff
@@ -534,6 +533,142 @@ var _ = Describe("ParseReference", func() {
...
+						topology.RabbitmqClusterReference{
+							Name:      existingRabbitMQCluster.Name,
+							Namespace: existingRabbitMQCluster.Namespace,
+						},
+						"allowed2",
+						"")
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+			When("requested namespace is not in topology-allowed-namespaces annotation", func() {
```
Zerpet, line 628: Same as previous comment 🙂

rabbitmqclient/cluster_reference_test.go
```diff
@@ -534,6 +533,142 @@ var _ = Describe("ParseReference", func() {
...
+							Namespace: existingRabbitMQCluster.Namespace,
+						},
+						"allowed3",
+						"")
+					Expect(err).To(MatchError(rabbitmqclient.ResourceNotAllowedError))
+				})
+			})
+		})
+
+		When("the rabbitmq allows all namespaces", func() {
```
Zerpet, line 642: RabbitMQ does not allow the namespaces, the Operator does 🙃 may I suggest "when all namespaces are allowed" ?

Or if you want a funny tweak "when your allow list is made of stars" 🤪

Zerpet (approved):
Left cosmetic suggestions, feel free to merge as-is 👍

coro (approved):

PR #604, opened 2023-05-02
Allow scoping cache to specific namespaces

PocketMobsters:

## Summary Of Changes

Adds an env variable `OPERATOR_SCOPE_NAMESPACE` that allows for scoping the Kubernetes client cache to one or more namespaces to allow for the operator to be utilized with non-cluster roles.

## Additional Context

If someone needed to configure their RBAC with non-cluster `roles` and `rolebindings` they would be unable to run the operator. This is due to the behavior of the Kubernetes client and how it handles caching. Without providing a namespace or a multinamespace cache, it'll default to caching at the cluster scope.

This PR adds behavior already existing in the core `cluster-operator` project, as well as an extension to that behavior I have PR'd in https://github.com/rabbitmq/cluster-operator/pull/1341 to allow for scoping the cache to multiple namespaces.

Without a lever such as `OPERATOR_SCOPE_NAMESPACE` and using `role` instead of `clusterrole` you get errors about being unable to list resources at the cluster scope. An example would be

```
pkg/mod/k8s.io/client-go@v0.26.3/tools/cache/reflector.go:169: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:rabbit:rabbit-cluster-operator" cannot list resource "secrets" in API group "" at the cluster scope
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,15 +12,17 @@ package main
 import (
 	"flag"
 	"fmt"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"os"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


Zerpet (approved):

Zerpet:
Verified this is working as documented in `controller-tools`. Started the manager with `OPERATOR_SCOPE_NAMESPACE` set to `rabbit-1,rabbit-2`. Objects created in any other namespace e.g. `default` are not reconciled, as expected. Objects created in either `rabbit-1` or `rabbit-2` are reconciled.

PR #630, opened 2023-05-31
Use golang 1.20

ChunyiLyu:
## Summary Of Changes

Use golang 1.20

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -12,7 +12,7 @@ on:
     tags: [ "v*" ]
 
 env:
-  GO_VERSION: ~1.19 # Require Go 1.19 and above, but lower than Go 2.0.0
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -12,7 +12,7 @@ on:
     tags: [ "v*" ]
 
 env:
-  GO_VERSION: ~1.19 # Require Go 1.19 and above, but lower than Go 2.0.0
+  GO_VERSION: ~1.20 # Require Go 1.20 and above, but lower than Go 2.0.0
```
MirahImage, line 15: This should already be using golang 1.20, as `~1.19` matches `1.20`

MirahImage (approved):
The only changes that are actually required to upgrade to 1.20 are in go.mod and the Dockerfile, as the other changes already matched 1.20. Nonetheless, it seems reasonable to update everywhere.

PR #633, opened 2023-06-09
Set status.cond when cluster secret not found

ChunyiLyu:
Reported in https://github.com/rabbitmq/messaging-topology-operator/discussions/612

## Summary Of Changes

Previously error was only logger, with this commit status.condition log and event are all updated.

Before:
```
Spec:
  Ack Mode:  on-confirm
  Name:      origin
  Rabbitmq Cluster Reference:
    Connection Secret:
      Name:  secret-that-doesnt-exist
  Uri Secret:
    Name:  federation-uri
  Vhost:   downstream
Events:    <none>
```

With this change:
```
Spec:
  Ack Mode:  on-confirm
  Name:      origin
  Rabbitmq Cluster Reference:
    Connection Secret:
      Name:  secret-that-doesnt-exist
  Uri Secret:
    Name:  federation-uri
  Vhost:   downstream
Status:
  Conditions:
    Last Transition Time:  2023-06-09T12:19:49Z
    Message:               failed to retrieve cluster from reference: Secret "secret-that-doesnt-exist" not found
    Reason:                FailedCreateOrUpdate
    Status:                False
    Type:                  Ready
Events:
  Type     Reason                Age                   From                   Message
  ----     ------                ----                  ----                   -------
  Warning  FailedCreateOrUpdate  65s (x15 over 2m27s)  federation-controller  failed to retrieve cluster from reference: Secret "secret-that-doesnt-exist" not found
```

### DIFF
--- a/controllers/topology_controller.go
+++ b/controllers/topology_controller.go
@@ -156,7 +156,17 @@ func (r *TopologyReconciler) handleRMQReferenceParseError(ctx context.Context, o
 		}
 		return ctrl.Result{}, nil
 	}
+
+	// set status condition and publish event for any other error
 	logger.Error(err, failedParseClusterRef)
+	msg := fmt.Sprintf("%s: %s", failedParseClusterRef, err.Error())
+	r.Recorder.Event(object, corev1.EventTypeWarning, "FailedCreateOrUpdate", msg)
+	object.SetStatusConditions([]topology.Condition{topology.NotReady(msg, r.getStatusConditions(object))})
+	if writerErr := clientretry.RetryOnConflict(clientretry.DefaultRetry, func() error {
+		return r.Status().Update(ctx, object)
+	}); writerErr != nil {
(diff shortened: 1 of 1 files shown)


controllers/topology_controller.go
```diff
@@ -156,7 +156,17 @@ func (r *TopologyReconciler) handleRMQReferenceParseError(ctx context.Context, o
 		}
 		return ctrl.Result{}, nil
 	}
+
+	// set status condition and publish event for any other error
 	logger.Error(err, failedParseClusterRef)
+	msg := fmt.Sprintf("%s: %s", failedParseClusterRef, err.Error())
+	r.Recorder.Event(object, corev1.EventTypeWarning, "FailedCreateOrUpdate", msg)
```
michaelklishin, line 163: Just FTR, I wanted to suggest using a "type error" here but apparently there only "normal" and "warning" to choose from in the API :)

michaelklishin (approved):

PR #644, opened 2023-06-27
Support custom connect uri through annotation on rabbitmqcluster

ChunyiLyu:
This closes #

## Summary Of Changes

Set ``


MirahImage (approved):
We should also document this, possibly with an example, but otherwise it looks good.

PR #682, opened 2023-09-21
Bumping kustomize and client go versions

DanielePalaia:
Same as https://github.com/rabbitmq/cluster-operator/pull/1436

Upgrading kustomize to V5 (which requires client-go v0.28.1) should solve issues like:
https://github.com/rabbitmq/messaging-topology-operator/actions/runs/6178266133/job/16771204981?pr=676


DanielePalaia:
Seems also that path: needs to be added before the actual path from kustomize v5

Issue #51, opened 2021-03-11
Scope operator interaction to namespace


Zerpet:
Looking at the `RabbitmqClusterReference` property in `User` kind:

https://github.com/rabbitmq/messaging-topology-operator/blob/06637f5b9fbacefce0ca2334577effd610c7a9eb/api/v1alpha1/user_types.go#L18-L33

https://github.com/rabbitmq/messaging-topology-operator/blob/06637f5b9fbacefce0ca2334577effd610c7a9eb/api/v1alpha1/queue_types.go#L44-L49

I noticed that we allow to interact with `RabbitmqCluster` objects in any namespace in the cluster. I'm concerned that we might be allowing a human user `Cody` to create RabbitMQ primitives in a RabbitMQ cluster they might not have access to. For example, `Cody` has access only to namespace `cody-dev` and they create a `User` kind which references a `RabbitmqCluster` with name `rabbit` in namespace `foobar`. The topology operator will create the user primitive in RabbitMQ, granting `Cody` credentials to access said RabbitMQ instance. 

My concern is that the operator might allow `Cody` to access RabbitMQ instances they are not supposed or intended to.

Should we remove the `namespace` aspect, effectively having a `LocalObjectReference`-like object?

MirahImage:
This is a breaking change to the API, therefore we need to bump the API version to v1alpha2 (issue #63)

coro:
Just to clarify, this affects all topology objects, which all allow for a namespace field in the cluster reference.

coro:
@n3wscott How would this proposed change affect knative?

PR #700, opened 2023-11-13
Fix issue with RabbitmqClusterReference Namespace Namespace and ConnectionSecret

### LINKED ISSUE #501: Connection Secret: Use from other namespaces
Managing an external, but centrally managed RabbitMQ cluster with applications in different namespaces (which should not be exposed to the admin credentials) is currently impossible.

The `connectionSecret` reference should allow setting the namespace of the connection secret. Additionally, restrictions on the namespaces allowed to do so should be possible (with the same annotation as the cluster object?).

Copying the connection secret to all application namespaces is possible, but highly undesirable (due to management issues and admin password exposure).

**Additional context**
#294  introduced the connection secret from #135.

mindw:
This closes #501

## Summary Of Changes
Fix an issue where Namespace was ignored when using ConnectionSecret.

## Additional Context
This doesn't look like a feature - it's more like a bug. Why support RabbitmqCluster via namespace but forbid secrets?

### DIFF
--- a/rabbitmqclient/cluster_reference.go
+++ b/rabbitmqclient/cluster_reference.go
@@ -35,19 +35,19 @@ var (
 )
 
 func ParseReference(ctx context.Context, c client.Client, rmq topology.RabbitmqClusterReference, requestNamespace string, clusterDomain string, connectUsingHTTP bool) (map[string]string, bool, error) {
-	if rmq.ConnectionSecret != nil {
-		secret := &corev1.Secret{}
-		if err := c.Get(ctx, types.NamespacedName{Namespace: requestNamespace, Name: rmq.ConnectionSecret.Name}, secret); err != nil {
-			return nil, false, err
-		}
-		return readCredentialsFromKubernetesSecret(secret)
-	}
-
 	var namespace string
 	if rmq.Namespace == "" {
 		namespace = requestNamespace
 	} else {
 		namespace = rmq.Namespace
+	}
+
+	if rmq.ConnectionSecret != nil {
+		secret := &corev1.Secret{}
+		if err := c.Get(ctx, types.NamespacedName{Namespace: namespace, Name: rmq.ConnectionSecret.Name}, secret); err != nil {
+			return nil, false, err
+		}
+		return readCredentialsFromKubernetesSecret(secret)
 	}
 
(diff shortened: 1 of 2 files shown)


Zerpet (changes requested):
We can't accept this contribution as it stands. For historical context see #51 #147 and #201.

The main reason to reject is the security implications this change may have. The user running the Topology Operator has access to a wide set of resources, secrets among them. The Topology Operator cannot know if user `Cody` has access to a namespace other than the namespace a request is coming from. For example: `Cody` creates a `Queue` in namespace `dev`. The existence of `Queue` object in NS `dev` is sufficient to conclude that `Cody` has access to NS `dev`, at least to create a Topology object. We can't know whether `Cody` has access to any other NS like `app-1`.

As illustrated in the example above, we would be allowing `Cody` to bypass NS restrictions, and therefore, we can't accept this as it currenty stands.

mindw:
@zerpet , thank you for the detailed and prompt reply.

From what I gather, it would be enough to add support for the `rabbitmq.com/topology-allowed-namespaces` annotation in the secret, with the same semantics as it has when used on a `RabbitmqClusterReference` object.

I'll update the PR and tests then.

Zerpet:
Bear in mind that when `ConnectionSecret` is used, there's not necessarily a `RabbitmqCluster` object. The `ConnectionSecret` was initially designed to use the Topology Operator with non-managed RabbitMQ servers [1]. Later we found some use cases where it makes sense to use `ConnectionSecret` in conjuction with `RabbitmqCluster` e.g. the creator of `RabbitmqCluster` overrides the `default_user` and `default_pass`.

The solution you propose will be acceptable in the specific scenario where a `RabbitmqCluster` object exists. Otherwise, the Secret **must** be in the same namespace.

[1] https://rabbitmq.com/kubernetes/operator/using-topology-operator.html#non-operator

mindw:
@zarpet - the annotation `rabbitmq.com/topology-allowed-namespaces` is now required on the referenced secret - same as it is on the `RabbitmqCluster`. This should satisfy the requirement.

Again, thank you the detailed explanations and your time!

Zerpet (approved):
Looks good, thank you!

Zerpet:
I will merge after CI passes

PR #710, opened 2023-12-01
Update controller-runtime to 0.16

Zerpet:

## Summary Of Changes

- Updated controller-runtime to 0.16.x
- Refactored Topology Controller tests
- Accepted changes suggested by GoLand code analysis tool
- Make the integration test suite run in parallel

## Additional Context

This version of controller runtime has breaking changes. Refactored the
Topology Controller tests because they were flaky. This group of tests
was defining their own queue reconciler in the same manager that has all
the controllers (included one for Queue types). This caused a "race"
between all setup reconcilers, and the test passed when the controller
registered in the specific test case was the first to reconcile, which
seemed to happen often enough to hide the flake.

After the refactor, each test case defines its own manager, which
watches a dedicated namespace for the Topology Controller tests, and
cancels/stops the manager at the end of each test case. This ensures
that only one controller is reconciling the Queue object. The
controllers defined in the BeforeSuite now watch only the default
namespace. All our tests were using the default namespace anyway, so
there are no failures as part of this subtle change.


Zerpet:
The SuperStream test suite is still flaking. Looking into it 👀

Zerpet:
This is ready for review after b8f7820 🥳

Zerpet:
Ah, I removed it locally because it was complaining about my local Go version and not being very useful at all. I'll add it back to the unit-tests target. IMO it's unnecessary to run them at unit **and** at integration level.

Issue #747, opened 2024-01-23
[Failsafe] Prevent topology operator to delete non-empty queues


jdfresser:
**Is your feature request related to a problem? Please describe.**
Sometimes, you could delete a queue.rabbitmq.com object by inadvertance.
For example, recently, we had an issue with our Gitops configuration that pruned everything in our clusters.
We easily recreated the topology, but the messages that where still inside the queues were lost.

**Describe the solution you'd like**
Implement a 'global' option `queue-delete-if-empty` in the topology operator.
In case of rabbitmq 'queue' objects deleted, the Queues will be deleted only if empty.

Rabbit-hole library already implements this [option](https://github.com/michaelklishin/rabbit-hole/blob/main/queues.go#L425C2-L425C9)
(Bonus: the topology operator can retry to delete it until queues are empty).

**Describe alternatives you've considered**
I've seen no similar Feature Request nor details about such a feature in the documentation.

Zerpet:
I'm happy to see this in the Topology Operator. It will be an opt-in feature, because changing the default behaviour would be a breaking change.

I'm not sure about the global option. I think this will be more interesting if implemented as a field in the `Queue` object. There are YAML manipulation tools like Kustomize or [YTT](https://carvel.dev/ytt/) to help with adding this new field to multiple YAML manifests.

jdfresser:
@Zerpet Agree with you, it must be an opt-in feature.

You're right, I was only considering my use-case, but it will be better to implement it as queue Queue spec item, so that users can define it by Queue  (It will also be easier to implement). . They should be mutable resources to allow users to use this feature whithout having to delete/recreate all queues. 

Should this enhancement also include the 'if-unused' option (i.e. do not delete the queue if it has consumers)?

One tricky point I didn't see at first is the incompatibility of if-empty flag with quorum queues:
```
{"error":"bad_request","reason":"cannot delete queue 'q.test' in vhost '/'. queue.delete operations with if-empty flag set are not supported by quorum queues"}
```
So, the topology operator should unset the flag if quorum queue and test the number of messages in queue before deleting it, I guess?

FYI, we do use Kustomize, but we rely on home-made python scripts to generate Topology config (and a nice mermaid schema for readability), so it'll be easy to add this field for us. We hope to simplify it by using Kluctl

MirahImage:
Closed by #759

PR #759, opened 2024-02-09
Allow usage of if-empty, if-unused parameters for queue deletion

jdfresser:
This closes #https://github.com/rabbitmq/messaging-topology-operator/issues/747

It probably misses some tests in https://github.com/rabbitmq/messaging-topology-operator/pull/759/files#diff-5619a2073d5cfff0b764ead1f1323f7554a40d748c906bcaa1322f1dd8a649dfR103

## Summary Of Changes
- Support 'failsafe' strategies to prevent the operator to delete queues with consumers/messages
- Modifies the API with two new 'mutables' fields in Queue CRD: Spec/DeleteIfEmpty & Spec/DeleteIfUnused
- For Quorum queues, check the rabbitMQ HTTP API to know if there are consumers/messages
- For other queues, use the built-in parameters

### DIFF
--- a/controllers/queue_controller.go
+++ b/controllers/queue_controller.go
@@ -47,11 +47,30 @@ func (r *QueueReconciler) DeleteFunc(ctx context.Context, client rabbitmqclient.
 	logger.Info("Deleting queues from ReconcilerFunc DeleteObj")
 
 	queue := obj.(*topology.Queue)
-	err := validateResponseForDeletion(client.DeleteQueue(queue.Spec.Vhost, queue.Spec.Name))
-	if errors.Is(err, NotFound) {
+	queueDeleteOptions, err := internal.GenerateQueueDeleteOptions(queue)
+	if err != nil {
+		return fmt.Errorf("failed to generate queue delete options: %w", err)
+	}
+	// Manage Quorum queue deletion if DeleteIfEmpty or DeleteIfUnused is true
+	if queue.Spec.Type == "quorum" && (queue.Spec.DeleteIfEmpty || queue.Spec.DeleteIfUnused) {
+		qInfo, err := client.GetQueue(queue.Spec.Vhost, queue.Spec.Name)
+		if err == nil {
+			return fmt.Errorf("cannot get %w Queue Information to verify queue is empty/unused: %s", err, queue.Spec.Name)
+		}
+		if qInfo.Messages > 0 && queue.Spec.DeleteIfEmpty {
+			return fmt.Errorf("Cannot delete queue %s because messages", queue.Spec.Name)
+		}
+		if qInfo.Consumers > 0 && queue.Spec.DeleteIfUnused {
+			return fmt.Errorf("Cannot delete queue %s because queue has consumers", queue.Spec.Name)
+		}
+
+	}
+
+	errdel := validateResponseForDeletion(client.DeleteQueue(queue.Spec.Vhost, queue.Spec.Name, *queueDeleteOptions))
+	if errors.Is(errdel, NotFound) {
 		logger.Info("cannot find queue in rabbitmq server; already deleted", "queue", queue.Spec.Name)
-	} else if err != nil {
-		return err
+	} else if errdel != nil {
+		return errdel
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


controllers/queue_controller.go
```diff
@@ -47,11 +47,30 @@ func (r *QueueReconciler) DeleteFunc(ctx context.Context, client rabbitmqclient.
...
+	queueDeleteOptions, err := internal.GenerateQueueDeleteOptions(queue)
+	if err != nil {
+		return fmt.Errorf("failed to generate queue delete options: %w", err)
+	}
+	// Manage Quorum queue deletion if DeleteIfEmpty or DeleteIfUnused is true
+	if queue.Spec.Type == "quorum" && (queue.Spec.DeleteIfEmpty || queue.Spec.DeleteIfUnused) {
+		qInfo, err := client.GetQueue(queue.Spec.Vhost, queue.Spec.Name)
+		if err == nil {
+			return fmt.Errorf("cannot get %w Queue Information to verify queue is empty/unused: %s", err, queue.Spec.Name)
+		}
```
Zerpet, line 59: Shouldn't this be `if err != nil` ?

controllers/queue_controller.go
```diff
@@ -47,11 +47,30 @@ func (r *QueueReconciler) DeleteFunc(ctx context.Context, client rabbitmqclient.
...
+		if qInfo.Messages > 0 && queue.Spec.DeleteIfEmpty {
+			return fmt.Errorf("Cannot delete queue %s because messages", queue.Spec.Name)
```
Zerpet, line 61: Error messages should not be capitalised. This would read better as `cannot delete queue %s because it has ready messages`

controllers/queue_controller.go
```diff
@@ -47,11 +47,30 @@ func (r *QueueReconciler) DeleteFunc(ctx context.Context, client rabbitmqclient.
...
+		}
+		if qInfo.Consumers > 0 && queue.Spec.DeleteIfUnused {
+			return fmt.Errorf("Cannot delete queue %s because queue has consumers", queue.Spec.Name)
```
Zerpet, line 64: Error message should not be capitalised.

system_tests/queue_system_test.go
```diff
@@ -100,5 +100,7 @@ var _ = Describe("Queue Controller", func() {
 			return err
 		}, 30).Should(HaveOccurred())
 		Expect(err.Error()).To(ContainSubstring("Object Not Found"))
+
+		//FIXME implement delete queue test with ifUnused/ if Empty?
```
Zerpet, line 104: I don't think we need to. We would be testing RabbitMQ API. From our point of view, it is sufficient to test that we send the adequate request to rabbit; then we rely on rabbit doing the right thing.

controllers/queue_controller.go
jdfresser: Yes, you're right!

internal/queue_delete_options.go
```diff
@@ -0,0 +1,34 @@
...
+
+// GenerateQueueDeleteOptions generates rabbithole.QueueDeleteOptions for a given Queue
+// queue.Spec.Arguments (type k8s runtime.RawExtensions) is unmarshalled
+func GenerateQueueDeleteOptions(q *topology.Queue) (*rabbithole.QueueDeleteOptions, error) {
+	arguments := make(map[string]interface{})
+	if q.Spec.Arguments != nil {
+		if err := json.Unmarshal(q.Spec.Arguments.Raw, &arguments); err != nil {
+			return nil, fmt.Errorf("failed to unmarshall queue arguments: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
jdfresser: My bad, I inspired myself from queue_settings.go whithout noticing I didn't need q.Spec.arguments

jdfresser:
@Zerpet, Thank you for your review, I applied the suggested changes. 

>  I want to check something with the team, because I'm unsure why Quorum Queues do not support if-empty and if-unused. Perhaps we could add that to the core, and that would simplify a lot this PR.

I'm also curious about the reason. Yes, better to avoid work-aroundsif we can!

Zerpet (approved):
Looking good, I'm going to do some manual QA before merging.

Zerpet:
> @Zerpet, Thank you for your review, I applied the suggested changes.
> 
> > I want to check something with the team, because I'm unsure why Quorum Queues do not support if-empty and if-unused. Perhaps we could add that to the core, and that would simplify a lot this PR.
> 
> I'm also curious about the reason. Yes, better to avoid work-aroundsif we can!

@jdfresser the main reason is that we can't deliver on either promise without some racy conditions. In QQ, a publisher can publish to any member (usually 3) and a consumer can consume from any replica. This fact makes it harder to deliver on "if empty" or "if unused" without some limitations. It' not the same case for classic queues, CQ can deliver both promises.

FYI I opened this issue to support "if empty" and "if unused" in QQs, but it won't make it until rabbit 4.0: https://github.com/rabbitmq/rabbitmq-server/issues/10543

PR #767, opened 2024-03-01
Testing olm

DanielePalaia:
Similar to https://github.com/rabbitmq/cluster-operator/pull/1569.

This PR is part of the task to migrate our Concourse pipelines to github actions:

It is migrating the OLM pipeline in order to test our OLM bundle in the messaging-topology-operator:
https://github.com/pivotal/rabbitmq-for-kubernetes-ci/blob/main/pipelines/olm-bundles/pipeline.yml

it can be useful to have a look to this guideline in order to see how to create and publish an OLM bundle:
https://devopstales.github.io/home/oml/

The test images are stored in the environment variable specified in:
 DOCKER_REGISTRY_SERVER: quay.io/rabbitmqoperator
 OLM_IMAGE: rabbitmq-for-kubernetes-olm-messaging-topology-operator
 OLM_INDEX_IMAGE: rabbitmq-for-kubernetes-olm-messaging-topology-operator-index


DanielePalaia:
The pipeline seems stable. I will merge it and test with the next tag release.

Zerpet:
This hasn't updated in a long time, but perhaps it still creates the bundle that we can use?

https://github.com/marketplace/actions/olm-bundle

DanielePalaia:
@Zerpet yes it is nicer to use checkout, I will fix it.

About this https://github.com/marketplace/actions/olm-bundle it is quite interesting.
I spent some time trying to find a script to do this conversion before implementing the simple script in: https://github.com/rabbitmq/OLM-Package-Repo
but I wasn't able to find anything except the script that operatorhub has in its own UI (without any reusable code).
I didn't think to search for a github-action.

If the action is working we can use directly that one avoiding to refactor and maintaining that script which is good.
Anyway this will take more testing. At this point I would prefer to do this testing eventually in a new iteration/PR as part of the issue we have to review this OLM part in general

DanielePalaia:
I replaced the git clones with the better checkout actions and I have tested it. I will now merge this one to test it on the next release. We can come back here when we will improve the OLM packaging script.

Issue #783, opened 2024-03-19
Slow memory leak


ollie-nye:
## Describe the bug

We're seeing a slow memory leak over the course of a few days that's resulting in the operator pod being OOMkilled. Logs look fairly normal except for a few recurring errors, all following these patterns:
- `failed to delete finalizer: Operation cannot be fulfilled on exchanges.rabbitmq.com ""<exchange_name>"": the object has been modified; please apply your changes to the latest version and try again`
- `failed to delete finalizer: Operation cannot be fulfilled on bindings.rabbitmq.com "<binding>": StorageError: invalid object, Code: 4, Key: /registry/rabbitmq.com/bindings/<binding>, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: <some uid>, UID in object meta: `

It's not much of a leak, just enough to start throwing alerts and then kill the pod every few days. (Usage screenshot below)

I couldn't see any existing issues for something related so opening up a new one

## To Reproduce
Our manifests are managed via Argo, but that shouldn't be getting involved in resources the operator looks after. Outside of that, we're running a standard manifest: https://github.com/rabbitmq/messaging-topology-operator/releases/download/v1.13.0/messaging-topology-operator-with-certmanager.yaml

We have a persisted CA being patched in with Kustomize but that doesn't look to be affecting finalizers.

**Expected behavior**
Operator memory usage to remain stable over longer periods of time

## Screenshots

![Screenshot 2024-03-13 at 08 59 57](https://github.com/rabbitmq/messaging-topology-operator/assets/26125311/802d49e1-a094-4ab2-83db-e099ea686ed4)

## Version and environment information
- Messaging Topology Operator: 1.13.0
- RabbitMQ: 3.13.0
- RabbitMQ Cluster Operator: 2.0.0
- Kubernetes: v1.27.8
- Cloud provider or hardware configuration: GKE
 
## Additional context

This has been over the last 30 days at least, I can't say when it started exactly because we don't keep metrics longer than that. We jumped from 1.10.3 to 1.13.0 and the alerts started shortly afterwards

Zerpet:
@mkuratczyk FYI this may be similar to this https://github.com/rabbitmq/cluster-operator/pull/1549

@ollie-nye how many `Secret` type objects do you have in your cluster? Does this number vary over time?

ollie-nye:
[...]

PR #808, opened 2024-05-07
Don't watch secrets

mkuratczyk:
Watching the secrets can lead to high memory usage if there's a lot of them. It seems like we don't actually rely on this, so we can turn that off completely.

Fixes https://github.com/rabbitmq/messaging-topology-operator/issues/783

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,14 +14,14 @@ import (
 	"fmt"
 	"os"
 	"regexp"
-	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"strconv"
 	"strings"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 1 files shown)


Zerpet (approved):
I think this will have the side effect that updating the `Secret` won't update the user credentials in RabbitMQ. I think it's a fair trade-off, because we are fixing a legitimate problem. As a workaround, folks can use time-based reconciliation.

mkuratczyk:
I just tried that with the 1.13.0 and it didn't work anyway. What I did was:
1. deploy the `userPreDefinedCreds.yaml` example
2. check with `rabbitmqctl authenticate_user`
3. change the secret and re-apply `userPreDefinedCreds.yaml`
4. check again with `rabbitmqctl authenticate_user` - the old password works, the new one doesn't

That's in line with the comment: https://github.com/rabbitmq/messaging-topology-operator/blob/main/docs/examples/users/userPreDefinedCreds.yaml#L7

Zerpet:
Sure, it won't work with pre-defined creds, but it should work with secrets created as part of the `User` creation. In any case, time-based reconciliation would be the workaround.

mkuratczyk:
This functionality still works for me:
1. deploy `docs/examples/users/user.yaml`
2. `rabbitmqctl authenticate_user`
3. edit the `user-sample-user-credentials`
4. try the new password

Seems like we didn't lose anything and we gained lower memory usage and higher security. :)

PR #847, opened 2024-07-01
WIP: Adding an action to keep versions/crds/clusterrole/webhook configurations in sync

DanielePalaia:
## Summary Of Changes
This action triggered on a new operator release will submit a PR to our helm-chart commercial repo in order to keep the image versions, the crds, the clusterrole and the webhook configurations updated.

It is using some internal python scripts in order to manage the update.


DanielePalaia:
This one will be replaced by an internal concourse pipeline

Issue #866, opened 2024-08-06
Support PasswordHash in User type


MirahImage:
RabbitMQ supports creating users with password hashes in addition to simply passwords. It would be useful if the messaging topology operator also supported this option for user creation. 

This could potentially also solve #865, as a passwordless user corresponds to an empty password hash.

Zerpet:
Fixed via #886

PR #886, opened 2024-09-14
Support PasswordHash in User type

### LINKED ISSUE #865: Creating a passwordless user
RabbitMQ allows creating [passwordless users](https://www.rabbitmq.com/docs/passwords#passwordless-users). However when an User resource is added via this operator, the password is required. If the password is not provided, it is automatically generated.

I see two possible solutions:
1. If the secret has no `password` key, generate the password; but if the secret has `password: ""`, disable password auth.  (Not backwards compatible)
2. Add a new `noPassword: bool` flag to UserSpec.

With the right configuration, password auth via AMQP can be disabled for all users, but not individually.

**Additional context**
[Management API](https://rawcdn.githack.com/rabbitmq/rabbitmq-server/v3.13.6/deps/rabbitmq_management/priv/www/api/index.html) endpoint `/api/users/*name*` creates a paswordless user if `password_hash: ""`.

I can implement the `noPassword: bool` solution and open a PR, just want to be sure that this isn't already possible, and that the maintainers are okay with this solution.

NikSays:
This closes #865, #866

## Summary Of Changes

The Secret specified in User's ImportCredentialsSecret may contain the `passwordHash` field.
If the field is absent, the behavior is unchanged: the plain-text password is generated if necessary and stored in the credentials Secret.

If the `passwordHash` field is present, then `password` field is ignored and the resulting credentials Secret will contain only the hash. This is done to prevent ambiguous situations where the hash doesn't correspond to the password. It also prevents the plain-text password being accidentally used over the hash.

If the hash is an empty string, a passwordless user is created.

## Additional Context

Currently the hash must be generated with the SHA512 algorithm. A way to specify another hashing algorithm needs to be added. I will add it as another field in the secret, but I'm concerned the documentation for importCredentialsSecret will become too complicated.

#### Todo

- [ ] ~Specify hashing algorithm~
- [x] Add examples
- [x] Update documentation


controllers/user_controller.go
```diff
@@ -40,6 +40,18 @@ const (
...
+// If the secret is not provided, a random username and password will be generated.
+type UserCredentials struct {
+	// Must be present if ImportCredentialsSecret is provided.
+	Username string
+	// If PasswordHash is an empty string, a passwordless user is created.
+	// If PasswordHash is nil, Password is used instead.
+	PasswordHash *string
+	// If Password is empty and PasswordHash is nil, a random password is generated.
+	Password string
+}
```
Zerpet, line 53: If this is not part of the API, I would prefer to move this struct to the `internal` package.

Zerpet (commented):
I had a quick look and it looks good overall. I want to review again the system test changes in more detail tomorrow.

Zerpet:
Thank you for contributing this feature. I like the high level logic proposed.

> [...]. A way to specify another hashing algorithm needs to be added. [...]

I'm not sure we need to, can you elaborate on this?

NikSays:
> I'm not sure we need to, can you elaborate on this?

I've realized that forcing SHA-512 is viable. It won't work only if someone needs to create the User with a SHA-256 hash without access to the original password, or if someone wants to use a custom hashing function provided by a plugin. Both scenarios seem unlikely. 
Implementing algorithm selection would require more effort, so until someone requests it or provides a use-case, SHA-512 should be enough.

NikSays:
I've added unit and system tests, updated the documentation, regenerated the asciidoc and the CRD. Also the branch was rebased on top of main, to simplify the commit history.

Zerpet (approved):

Zerpet:
The latest changes look good to me. I'll merge after CI passes 🚀

NikSays:
Hi, @Zerpet, sorry to bother you. Could you please restart the CI workflow, there was a timeout starting the cluster operator, seems like a one-off issue.

Zerpet:
Thank you!

PR #896, opened 2024-10-31
Regenerate user passwords when user reconciles

### LINKED ISSUE #242: Update user info if targetsecret changes
deploy the user example. then update the password and deploy again. The User is not reconciled with the new password. 
This will make it easy for operators to rotate password or fix username typos.

When a users original secret is changed it should be reconciled with the new info
````
  importCredentialsSecret:
    name: rabbit-test-user
````

Martinif:
This closes #242 

## Summary Of Changes
This Pull Request adds an option to always regenerate a user's user-credentials secret when a reconciliation for the user is triggered. The main use case is to update a user's password when the underlying import secret has changed.

## Additional Context
A user's login credential are stored in the user-credentials secrets, which are generated at the user creation from an import secret. The documentation instructs us to add a label to trigger user reconciliation when we want to update a users password: https://github.com/rabbitmq/messaging-topology-operator/blob/main/docs/examples/users/README.md

This does however currently not affect a regeneration of the user-credentials secret: s. issue #242 

Therefore I added these changes to allow us to force a regeneration when the user is reconciled (e.g. when a label is added).

### DIFF
--- a/api/v1beta1/user_types.go
+++ b/api/v1beta1/user_types.go
@@ -37,6 +37,10 @@ type UserSpec struct {
 	//
 	// Note that this import only occurs at creation time, and is ignored once a password has been set on a User.
 	ImportCredentialsSecret *corev1.LocalObjectReference `json:"importCredentialsSecret,omitempty"`
+	// Feature flag to always regenerate the `-user-credentials` Secret from the ImportCredentialsSecret.
+	// Defaults to false if omitted.
+	// +kubebuilder:validation:Optional
+	AutoUpdateCredentialsSecret bool `json:"autoUpdateCredentialsSecret,omitempty"`
 }
 
 // UserStatus defines the observed state of User.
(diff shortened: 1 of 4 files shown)


Zerpet (changes requested):
This approach is not going to work. In 9ce8ed9134560a1ba5ba9e4b3b9310022fe9e200 we removed the RBAC to `update` secrets, because the Operator was too privileged: it was able to read/list/create/update any secret in any namespace. We removed its ability to `update` secrets at cluster level, accepting the limitation that the Operator won't be able to update secrets it had created.

Another blocker to accept this contribution is the lack of tests. I caught the above mentioned problem by doing manual QA, which is not desirable.

The most we are willing accept is updating the password if the underlying "user-credentials" Secret changes, taking into consideration the points I raised here: https://github.com/rabbitmq/messaging-topology-operator/issues/242#issuecomment-930363766

Zerpet:
This PR hasn't seen any activity since October 31st. I'm closing it as stale. Feel free to re-open if there's still interest in driving this PR.

PR #935, opened 2025-01-17
Update deps

Zerpet:
## Summary Of Changes

- Update dependencies

## Additional Context

Dependabot is not cleaning the go sum correctly. I had to run `go mod tidy` after updating
all the dependencies.


Zerpet:
CI failed with the same error that we hit every time we update the cluster operator: hard-coded path cannot be found. I added a commit to use a glob pattern instead. Locally, the glob may return multiple paths, but that's fine because it is the same API version of the operator `v1beta1`. In CI, the glob will return just one path.

Zerpet:
@MirahImage @mkuratczyk anyone fancy a review? 👀

MirahImage (approved):

PR #962, opened 2025-03-18
Add user limits.

### LINKED ISSUE #960: User limits max-connections
I want to define user limit max-connections as described in: https://www.rabbitmq.com/docs/user-limits#connections

I would like to use the messaging-topology-operator for defining user max-connections limits.

We could use custom scripts to achieve this, but it would be so nice to have this converted in this operator.

**Additional context**
It would be best to have additional resource like Limits with the possibility to set max-connections / max-queues for a specific user / vhost. Basically what is supported on Admin -> Limits page on webui.

MirahImage:
This closes #960

## Summary Of Changes

Adds optional Limits for connections and channels to the user spec.


jaroslaw-bochniak (approved):

controllers/user_controller_test.go
```diff
@@ -154,6 +156,53 @@ var _ = Describe("UserController", func() {
...
+					})))
+				By("calling PutUserLimits with the correct user limits")
+				Expect(fakeRabbitMQClient.PutUserLimitsCallCount()).To(BeNumerically(">", 0))
+				_, userLimitsValues := fakeRabbitMQClient.PutUserLimitsArgsForCall(0)
+				connectionLimit, ok := userLimitsValues["max-connections"]
+				Expect(ok).To(BeTrue())
+				Expect(connectionLimit).To(Equal(5))
+				channelLimit, ok := userLimitsValues["max-channels"]
+				Expect(ok).To(BeTrue())
+				Expect(channelLimit).To(Equal(10))
```
Zerpet, line 203: You can replace these lines with matches `ToHaveKeyWithValue()`. It's a bit easier to understand what's going, and in the event of a failure, you get something better than `expected false to be true` and `123 not equal 5`
https://onsi.github.io/gomega/#havekeywithvaluekey-any-value-any

internal/user.go
```diff
@@ -62,3 +62,14 @@ func GenerateUserSettings(credentials *corev1.Secret, tags []topology.UserTag) (
...
 }
+
+func GenerateUserLimits(userLimits topology.UserLimits) rabbithole.UserLimitsValues {
+	userLimitsValues := rabbithole.UserLimitsValues{}
+	if userLimits.Connections > 0 {
+		userLimitsValues["max-connections"] = int(userLimits.Connections)
+	}
+	if userLimits.Channels > 0 {
+		userLimitsValues["max-channels"] = int(userLimits.Channels)
+	}
```
Zerpet, line 73: Zero is a valid value in RabbitMQ. Negative values are valid too. If we want to enforce this lower bound for values, I would enforce it with validation in the CRD, instead of silently ignoring the values here.

internal/user_test.go
```diff
@@ -63,4 +63,31 @@ var _ = Describe("GenerateUserSettings", func() {
...
+			Expect(limits["max-channels"]).To(Equal(7))
+		})
+
+		It("does not create unspecified limits", func() {
+			limits := internal.GenerateUserLimits(topology.UserLimits{
+				Connections: 5,
+			})
+			Expect(limits["max-connections"]).To(Equal(5))
+			_, ok := limits["max-channels"]
+			Expect(ok).To(BeFalse())
```
Zerpet, line 83: I would change these with `HaveKeyWithValue()` and `ToNot(HaveKey())`. Better readability and more intuitive assertion failures.

https://onsi.github.io/gomega/#havekeykey-any

system_tests/user_system_test.go
```diff
@@ -388,4 +392,82 @@ var _ = Describe("Users", func() {
...
+				return err
+			}, 30, 2).Should(BeNil())
+			Expect(userLimitsInfo).To(HaveLen(1))
+			Expect(userLimitsInfo[0].User).To(Equal(username))
+			connectionLimit, ok := userLimitsInfo[0].Value["max-connections"]
+			Expect(ok).To(BeTrue())
+			Expect(connectionLimit).To(Equal(4))
+			channelLimit, ok := userLimitsInfo[0].Value["max-channels"]
+			Expect(ok).To(BeTrue())
+			Expect(channelLimit).To(Equal(6))
```
Zerpet, line 454: Same here, I would change these to use matchers `HaveKeyWithValue()`

Zerpet (commented):
I'm curious about clearing the user limits. It looks like (I haven't tried) if a user limit is set, and then the `UserLimits` field is set to an empty struct, the declare logic would simply ignore the limits, because the zero value of `int32` is `0` and our logic ignores any values equal or below zero.

One option would be to set the `UserLimit` field as a pointer, so that we can differentiate between the zero value and not set at all (`nil`). Note that `0` for limits is a valid value in RabbitMQ.

internal/user.go
MirahImage: Yeah, I did something slightly different in the vhost limits, which I might move over to here. It's a bit more annoying, since it requires using nested pointers, but it solves the problem of distinguishing between unset and 0.

MirahImage:
@Zerpet it now handles the 0 and negative cases, as well as the change of limits.

PR #977, opened 2025-04-08
Add vhost limits.

### LINKED ISSUE #963: Add vhost limits
RabbitMQ supports [vhost limits](https://www.rabbitmq.com/docs/vhosts#limits). It would be nice to configure these limits via the messaging topology operator.

An optional additional Limits section in the vhost spec to specify limits. This is akin to the user limits added in #962

MirahImage:
This closes #963


jaroslaw-bochniak (approved):

system_tests/vhost_system_test.go
```diff
@@ -139,4 +139,61 @@ var _ = Describe("vhost", func() {
...
+			Expect(len(vhostLimitsInfoResponse)).To(Equal(1))
+			Expect(vhostLimitsInfoResponse[0].Vhost).To(Equal(vhostWithLimits.Spec.Name))
+			vhostLimitsValues := vhostLimitsInfoResponse[0].Value
+			Expect(len(vhostLimitsValues)).To(Equal(2))
+			connectionLimit, ok := vhostLimitsValues["max-connections"]
+			Expect(ok).To(BeTrue())
+			Expect(connectionLimit).To(Equal(int(connections)))
+			queueLimit, ok := vhostLimitsValues["max-queues"]
+			Expect(ok).To(BeTrue())
+			Expect(queueLimit).To(Equal(int(queues)))
```
Zerpet, line 196: Here definitely preferable to use `HaveKeyWithValue()`. The assertion error when the key does not exist is `expected false to be true`, which give no hint as to what went wrong.

Zerpet (commented):
I'm wondering about clearing the vHost limits. The field in the spec is a pointer, so perhaps we could clear the limits if the field is `nil` ?

As the code stands, once a value is set, it can be lifted by setting to a negative value, but it can't be cleared.

MirahImage:
Vhost limits can now be updated.

PR #1000, opened 2025-06-23
Fix crd-ref-docs workflow

Zerpet:
It's been broken for a while. Tested the fix locally using nektos/act.


MirahImage (approved):
It's interesting that it often worked on the second try, unclear to me why it worked at all.

PR #1164, opened 2026-04-30
Bump golang version.

MirahImage:


Zerpet (approved):
Makes me wonder if we should change the default of Durable, so that it is true by default

PR #1176, opened 2026-05-13
Fix User deletion logic

### LINKED ISSUE #1171: Cannot delete a User with an invalid importCredentialsSecret
### Describe the bug

If a User is created with an invalid "importCredentialsSecret" name, the resource will be stuck forever as the operator can never deletes it.

### Reproduction steps

1. Apply the following:
```Yaml
kind: User
metadata:
  name: an-user
spec:
  rabbitmqClusterReference:
    name: rabbitmq
  importCredentialsSecret:
    name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

2. Run `kubectl get user.rabbitmq.com/an-user -o yaml`
```
(...)
status:
  conditions:
  - lastTransitionTime: "2026-05-08T11:17:32Z"
    message: 'could not find password secret aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
      in namespace xxxxxxxx; Err: Secret "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
      not found'
    reason: FailedCreateOrUpdate
    status: "False"
    type: Ready
```

3. Try to delete the user: `kubectl delete user.rabbitmq.com/an-user`
=> The command will be stuck
Even with --force nothing will change

The only way to remove it is to run `kubectl patch user.rabbitmq.com/an-user -p '{"metadata":{"finalizers":[]}}' --type merge`.

### Expected behavior

The deletion should not be stuck if the created ressource is invalid

### Additional context

_No response_

Zerpet:
This closes #1171

## Summary Of Changes

Fix user deletion logic.

## Additional Context

When a User resource fails to be created in RabbitMQ (e.g., due to missing
credentials secret), its status.username remains empty. Previously, attempting
to delete such a resource would try to delete a user with an empty username
from RabbitMQ, or get stuck in reconciliation.

This change adds a check in the user deletion handler to skip the RabbitMQ
deletion call when status.username is empty, allowing the Kubernetes resource
to be cleaned up properly. It also updates the permission controller to handle
references to users that don't yet have a username set, returning the user
object instead of an error to allow graceful handling downstream.

Includes test coverage for deleting a user that was never successfully created.

### DIFF
--- a/internal/controller/permission_controller.go
+++ b/internal/controller/permission_controller.go
@@ -72,9 +72,8 @@ func getUsernameFromUser(ctx context.Context, k8sClient client.Client, namespace
 
 	// get username from User status
 	if user.Status.Username == "" {
-		err := fmt.Errorf("this User does not have an username set in its status")
-		logger.Error(err, failureMsg, "userReference", name)
-		return nil, err
+		logger.Info("this User does not have a username set in its status", "userReference", name)
+		return user, nil
 	}
 	return user, nil
 }
--- a/internal/controller/user_controller.go
+++ b/internal/controller/user_controller.go
@@ -289,6 +289,11 @@ func (r *UserReconciler) DeleteFunc(ctx context.Context, rmqc rabbitmqclient.Cli
 	logger := ctrl.LoggerFrom(ctx)
 	user := obj.(*topology.User)
 
+	if user.Status.Username == "" {
+		logger.Info("user never successfully created in rabbitmq server; skipping deletion", "user", user.Name)
+		return nil
+	}
+
 	err := validateResponseForDeletion(rmqc.DeleteUser(user.Status.Username))
 	if errors.Is(err, ErrNotFound) {
 		logger.Info("cannot find user in rabbitmq server; already deleted", "user", user.Name)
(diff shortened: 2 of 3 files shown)


MirahImage:
Is there any way a user could get created in RabbitMQ and the username could remain unset in the status (for example, due to network problems)? Not suggesting we really need to guard against every case, just worried about potentially accidentally introducing a new edge case.

Otherwise, looks good.

Zerpet:
I don't see how that would be possible. Since we rely on having the `Status.Username` set to determine whether to declare credentials or not, if the Secret is declared and the Status remains empty (e.g. k8s API returns a conflict error), the call to rabbit API to `PUT` a user won't execute. Therefore, on next reconcile, the logic to generate creds will execute again, update the Secret credentials and set the status, only then the user will be created in rabbit.

I'm also reading that even in the event that rabbit API fails to create the user, the next reconcile will fetch existing credentials and attempt to create the user in rabbit.

All in all, I think the `User` controller logic is very resilient 😄

MirahImage (approved):

PR #1208, opened 2026-07-02
Mount cert-manager cert for serving metrics.

MirahImage:
The metrics endpoint previously served with controller-runtime's in-memory self-signed certificate: the cert-manager `metrics-server-cert` was created but never mounted (cert_metrics_manager_patch.yaml was commented out), and the e2e suite only asserted reachability via `curl -k`.

Note that topology operator secrets will continue to have generic names to aviod breaking changes, but will not collide with the cluster operator secrets which have unique names as of
https://github.com/rabbitmq/cluster-operator/pull/2189

[ai-assisted=yes]


Zerpet (approved):
I usually prefer more `ytt` and less `kustomize`, but I understand that agents are probably better at `kustomize` since `ytt` is "exotic"

PR #1209, opened 2026-07-03
RMQ-2841: Implement idempotent password update logic for User

Zerpet:
## Summary

Closes RMQ-2841. Builds on RMQ-2838's label-gated Secret cache/import-secret watch.

Today, `UserReconciler.DeclareFunc` calls `PutUser` on every successful reconcile, and `internal.GenerateUserSettings` salts the password hash freshly each call — so RabbitMQ receives a real credential-changing write on every reconcile, even when the password hasn't changed.

- Track `UserStatus.ObservedSecretVersion` (the credentials Secret's `resourceVersion` last successfully applied) alongside the existing `ObservedGeneration`, and skip `PutUser` when neither has changed since the last successful sync.
- Add a `PasswordSynced` condition (upserted alongside `Ready` — `User.SetStatusConditions` now merges by condition type instead of overwriting the whole slice).
- Emit a `Normal PasswordUpdated` Kubernetes Event on real syncs, via a new `UserReconciler.Recorder`.
- Fix a stale `ImportCredentialsSecret` doc comment claiming import only happens at creation time (no longer true since RMQ-2838).

**Scope tradeoff (intentional):** periodic resyncs no longer correct password drift made directly against RabbitMQ (e.g. via `rabbitmqctl`) — only k8s-triggered changes (Secret rotation, Tags edits) now re-sync. This matches the ticket's AC1 wording ("reconciliation triggered by a Secret change").

## Test plan

- [x] `go build ./...` / `go vet ./...` clean
- [x] `make generate manifests` — CRD/deepcopy diff reviewed, contains only the new `observedSecretVersion` field and the updated doc comment
- [x] `make just-unit-tests` — all 8 suites pass
- [x] `make just-integration-tests` — full 94/94 controller specs pass (confirms the `SetStatusConditions` change is a no-op for the other 11 CRs)
- [x] New/extended tests in `internal/controller/user_controller_test.go`:
  - no-op reconcile (metadata-only User change) doesn't re-call `PutUser`, for both imported- and generated-credentials paths
  - Tags-only spec change still triggers `PutUser`
  - real password rotation sets `PasswordSynced: True`, emits `Normal PasswordUpdated`, and leaves `Tags` unchanged in the `PutUser` payload

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -6,6 +6,7 @@ import (
 )
 
 const ready ConditionType = "Ready"
+const passwordSynced ConditionType = "PasswordSynced"
 
 type ConditionType string
 
@@ -24,7 +25,7 @@ type Condition struct {
 
 // Ready indicates that the last Create/Update operator on the CR was successful.
 func Ready(lastConditions []Condition) Condition {
-	time := lastTransitionTime(corev1.ConditionTrue, lastConditions)
+	time := lastTransitionTime(ready, corev1.ConditionTrue, lastConditions)
 	return Condition{
 		Type:               ready,
 		Status:             corev1.ConditionTrue,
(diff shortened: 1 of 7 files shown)


Zerpet:
Closing as per discussion with @MirahImage. We don't want the trade-off of password drift

PR #1210, opened 2026-07-03
RMQ-2841: Emit password sync event and condition for User

Zerpet:
## Summary

Closes RMQ-2841. Supersedes #1209, which proposed skipping `PutUser` when neither the User's generation nor the credentials Secret's `resourceVersion` had changed. That approach was rejected after discussion: it would have silently dropped the operator's ability to self-heal RabbitMQ-side password drift (e.g. a password changed directly via `rabbitmqctl`) on periodic resync.

This PR keeps the parts of that work that are useful independent of the skip-logic trade-off:

- `PutUser` continues to run on every reconcile, exactly as before this ticket — no skip logic, no drift-correction trade-off.
- Emit a `Normal PasswordUpdated` Kubernetes Event whenever `PutUser` succeeds, for both the imported- and generated-credentials paths, via a new `UserReconciler.Recorder`.
- Add a `PasswordSynced` condition set alongside `Ready` after a successful sync. `User.SetStatusConditions` now upserts by condition type instead of overwriting the whole slice, so the two conditions can coexist (verified as a no-op for the other 11 CRs, which only ever set a single condition type).
- Fix a stale `ImportCredentialsSecret` doc comment claiming import "only occurs at creation time" — not true since RMQ-2838 added password-rotation support via the import secret watch.

## Test plan

- [x] `go build ./...` / `go vet ./...` clean
- [x] `make generate manifests` — diff contains only the doc-comment fix, no new status fields
- [x] `make just-unit-tests` — all 8 suites pass
- [x] `make just-integration-tests` — full 91/91 controller specs pass (confirms the `SetStatusConditions` change is a no-op for the other 11 CRs)
- [x] New tests in `internal/controller/user_controller_test.go` assert `PasswordSynced: True` and a `PasswordUpdated` event after the first sync, and again after rotating the import secret's password

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/system/user_system_test.go
+++ b/test/system/user_system_test.go
@@ -111,8 +111,8 @@ var _ = Describe("Users", func() {
 				return updatedUser.Status.Conditions
 			}, waitUpdatedStatusCondition, 2).Should(HaveLen(2), "User status condition should be present")
 
-			Ω(updatedUser.Status.Conditions).Should(ContainElement(SatisfyAll(
-				HaveField("Type", "Ready"),
+			Expect(updatedUser.Status.Conditions).To(ContainElement(SatisfyAll(
+				HaveField("Type", topology.ConditionType("Ready")),
 				HaveField("Status", corev1.ConditionTrue),
 				HaveField("Reason", "SuccessfulCreateOrUpdate"),
 				HaveField("LastTransitionTime", Not(Equal(metav1.Time{}))),
--- a/test/system/user_system_test.go
+++ b/test/system/user_system_test.go
@@ -109,13 +109,14 @@ var _ = Describe("Users", func() {
 			Eventually(func() []topology.Condition {
 				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: user.Name, Namespace: user.Namespace}, &updatedUser)).To(Succeed())
 				return updatedUser.Status.Conditions
-			}, waitUpdatedStatusCondition, 2).Should(HaveLen(1), "User status condition should be present")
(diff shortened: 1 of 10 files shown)


test/system/user_system_test.go
```diff
@@ -109,13 +109,14 @@ var _ = Describe("Users", func() {
...
-			}, waitUpdatedStatusCondition, 2).Should(HaveLen(1), "User status condition should be present")
-
-			readyCondition := updatedUser.Status.Conditions[0]
-			Expect(string(readyCondition.Type)).To(Equal("Ready"))
-			Expect(readyCondition.Status).To(Equal(corev1.ConditionTrue))
-			Expect(readyCondition.Reason).To(Equal("SuccessfulCreateOrUpdate"))
-			Expect(readyCondition.LastTransitionTime).NotTo(Equal(metav1.Time{}))
+			}, waitUpdatedStatusCondition, 2).Should(HaveLen(2), "User status condition should be present")
+
+			Ω(updatedUser.Status.Conditions).Should(ContainElement(SatisfyAll(
```
MirahImage: Stylistically a bit of an odd choice to use Omega instead of Expect

MirahImage (changes requested):

test/system/user_system_test.go
Zerpet: The idea was to read: updated user status conditions Should contain an element that ...

I think it reads similarly but shorter than: Expect updated user status conditions To contain an element ...

That said, I don't have a strong opinion on that.

MirahImage (approved):

PR #1213, opened 2026-07-08
Add defaulting webhook to ensure default values.

### LINKED ISSUE #1172: Update queue defaults
### Is your feature request related to a problem? Please describe.

Starting with RabbitMQ 4, there have been some changes in behavior around durable, auto-delete, and exclusive queue properties to ensure that contradictory combinations are not used together. However, the default behavior of queue objects does not respect this.

### Describe the solution you'd like

Update the queue defaults to ensure that queues are durable, non-exclusive, and not auto-delete by default. 

We should carefully consider how this defaulting takes place to:
1. Avoid changing properties of existing queues (this is of course not possible on the RabbitMQ side, but the endlessly failing reconciliation loop would be bad)
2. Ensure that the way the defaulting occurs does not interfere with objects deployed by helm and similar methods. 

Using a webhook for defaulting would likely solve both of these problems. 

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

MirahImage:
Defautls Durable=true and autoDelete=false.

Durable internally changed to *bool to facilitate distinguishing false and nil values.

Because the defaulting webhook only impacts existing objects, this should not cause any changes to existing queues. This change of defaults could be seen as a breaking change, but the old defaults resulted in a failure to create with RabbitMQ 4.0+, so the change is necessary.

Resolves #1172

[ai-assisted=yes]


api/v1beta1/queue_types_test.go
```diff
@@ -79,7 +79,7 @@ var _ = Describe("Queue spec", func() {
 		Expect(fetchedQ.Spec.Name).To(Equal("test-queue"))
 		Expect(fetchedQ.Spec.Vhost).To(Equal("/hello"))
 		Expect(fetchedQ.Spec.Type).To(Equal("a type"))
-		Expect(fetchedQ.Spec.Durable).To(BeTrue())
+		Expect(ptr.Deref(fetchedQ.Spec.Durable, false)).To(BeTrue())
```
Zerpet: I think this assertion reads better if we use `PointTo` [1] or `HaveValue` [2]

[1] https://onsi.github.io/gomega/#testing-pointer-values
[2] https://onsi.github.io/gomega/#havevaluematcher-typesgomegamatcher

internal/managedresource/superstream_partition_test.go
```diff
@@ -58,7 +59,7 @@ var _ = Describe("SuperstreamPartition", func() {
 		})
 
 		It("sets the queue to be durable", func() {
-			Expect(partition.Spec.Durable).To(BeTrue())
+			Expect(ptr.Deref(partition.Spec.Durable, false)).To(BeTrue())
```
Zerpet: I think it reads better `To(PointTo(BeTrue()))` or `To(HaveValue(true))`

internal/webhook/v1beta1/webhook_suite_test.go
```diff
@@ -127,6 +127,9 @@ var _ = BeforeSuite(func() {
 	err = SetupQueueWebhookWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = SetupQueueDefaulterWebhookWithManager(mgr)
+	Expect(err).NotTo(HaveOccurred())
```
Zerpet:
```suggestion
	Expect(SetupQueueDefaulterWebhookWithManager(mgr)).To(Succeed())
```

1 line better than 2? :P

api/v1beta1/queue_types_test.go
MirahImage: Yeah, it actually was that way earlier, but for some reason when fixing a different issue with pointers, claude changed it to this awful syntax. I'll change it back.

Zerpet (approved):

PR #1247, opened 2026-08-17
Run OLM tests on PR

Zerpet:
Add a dedicated workflow to run OLM tests on every PR. I decided to use a separate workflow because PRs from external contributors do not have access to secrets, and the "release" OLM workflow pushes to Quay IO registry. This new workflow configures a local registry, and configures KinD to use this registry as mirror. This allows to run the package generation and system tests on _any_ PR.

### DIFF
--- a/.github/workflows/olm-pr.yml
+++ b/.github/workflows/olm-pr.yml
@@ -1,0 +1,128 @@
+---
+name: Test OLM bundle (PR)
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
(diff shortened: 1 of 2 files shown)


Zerpet:
Validated in CI: https://github.com/rabbitmq/messaging-topology-operator/actions/runs/32128632629/job/95684656114?pr=1247

MirahImage (approved):

Zerpet:
Not adding milestone because this is only a CI change.

